PR #11, opened 2014-08-15
Added ability for parse() to handle string booleans

apergy:


gapipro:
Maybe this should be an optional parameter to cast strings to booleans/numbers?

rgraffbrd:
@sindresorhus Yeah, so if you're going to reject stuff like this, you need to expose a means by which to add functionality like this without forking, y'know!

PR #17, opened 2015-03-31
fix isArray

Nemo64:
Array.isArray is not available in IE8.
https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_Unterst.C3.BCtzung

### DIFF
--- a/query-string.js
+++ b/query-string.js
@@ -32,7 +32,7 @@
 
 			if (!ret.hasOwnProperty(key)) {
 				ret[key] = val;
-			} else if (Array.isArray(ret[key])) {
+			} else if (ret[key] instanceof ret[key]) {
 				ret[key].push(val);
 			} else {
 				ret[key] = [ret[key], val];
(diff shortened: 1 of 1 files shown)


query-string.js
```diff
@@ -32,7 +32,7 @@
 
 			if (!ret.hasOwnProperty(key)) {
 				ret[key] = val;
-			} else if (Array.isArray(ret[key])) {
+			} else if (ret[key] instanceof ret[key]) {
```
kevva: `ret[key] instanceof Array)`?

kevva:
Not sure IE8 support is needed to be honest.

sindresorhus:
No thanks

Nemo64:
No thanks because you don't want ie8 support or no thanks because my
implementation is rushed because I quickly needed it?

Sindre Sorhus <notifications@github.com> schrieb am Di., 31. März 2015
17:07:

kevva:
No thanks to IE8 support.

PR #22, opened 2015-05-07
Optionally sort param keys in .stringify

pumpikano:
This is useful for my purposes at the moment to, e.g., test if two requests are really the same. Feel free to close this PR if you don't think this use case is general enough to include.

Thanks for the package!

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,21 +33,16 @@ exports.parse = function (str) {
 	}, {});
 };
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
We can just sort by default, no need for an option. Query strings are unordered. Just use https://github.com/sindresorhus/sort-keys

pumpikano:
Yup, that makes sense to me.

PR #23, opened 2015-05-23
Remove ampersand

saromanov:
Hi!
If query string starts with &, result from parse will be { '': null, data: 'item' }
```javascript
console.log(query.parse("&data=item")); //{ '': null, data: 'item' }
```
So, I've just a made more predictable behavior.
```javascript
console.log(query.parse("&data=item")); //{ data: 'item' }
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,10 +5,10 @@ exports.parse = function (str) {
 		return {};
 	}
 
-	if (str[0] == '&') {
+	if (str[0] === '&') {
 		str = str.substr(1, str.length);
 	}
-	
+
 	str = str.trim().replace(/^(\?|#)/, '');
 
 	if (!str) {
--- a/index.js
+++ b/index.js
@@ -11,6 +11,10 @@ exports.parse = function (str) {
 		return {};
 	}
 
+	if (str[0] == '&') {
+		str = str.substr(1, str.length);
+	}
+
 	return str.trim().split('&').reduce(function (ret, param) {
 		var parts = param.replace(/\+/g, ' ').split('=');
 		var key = parts[0];
--- a/index.js
+++ b/index.js
@@ -5,11 +5,7 @@ exports.parse = function (str) {
 		return {};
 	}
 
-	if (str[0] === '&') {
-		str = str.substr(1, str.length);
-	}
-
-	str = str.trim().replace(/^(\?|#)/, '');
+	str = str.trim().replace(/^(\?|#|&)/, '');
 
 	if (!str) {
 		return {};
--- a/index.js
+++ b/index.js
@@ -5,14 +5,14 @@ exports.parse = function (str) {
 		return {};
 	}
 
+	if (str[0] == '&') {
+		str = str.substr(1, str.length);
+	}
+	
 	str = str.trim().replace(/^(\?|#)/, '');
 
 	if (!str) {
 		return {};
-	}
-
-	if (str[0] == '&') {
-		str = str.substr(1, str.length);
 	}
 
 	return str.trim().split('&').reduce(function (ret, param) {
--- a/test.js
+++ b/test.js
@@ -11,7 +11,7 @@ describe('.parse()', function () {
 		assert.deepEqual(qs.parse('#foo=bar'), {foo: 'bar'});
 	});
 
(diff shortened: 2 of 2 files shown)


sindresorhus:
That's a invalid query string. The ampersand `&` is a separator. Why do you need this?

index.js
```diff
@@ -11,6 +11,10 @@ exports.parse = function (str) {
 		return {};
 	}
 
+	if (str[0] == '&') {
+		str = str.substr(1, str.length);
+	}
```
sindresorhus: Should implemented on line 8.

test.js
```diff
@@ -38,6 +38,10 @@ describe('.parse()', function () {
 	it('handle multiple of the same key', function () {
 		assert.deepEqual(qs.parse('foo=bar&foo=baz'), {foo: ['bar', 'baz']});
 	});
+
+	it('remove & from starts', function () {
```
sindresorhus:
```
it('query strings starting with a `&`'', function () {
```

and should be added on line 14 instead.

saromanov:
This is remove only from start of the string. Not from all of this. For example:
```javascript
console.log(query.parse("&data=foo&data=bar&fun=bag)); //{ data: [ 'foo', 'bar' ], fun: 'bag' }
```

sindresorhus:
No, I mean, why does your querystring start with a `&`? As I said, that's invalid, some I'm curious why it's like that.

saromanov:
Yes, i know that this is invalid, but my main motivation was replace incorrect behavior(in my point of view) with ampersand. In other cases, even if you put something incorrect, your response will be also invalid
```javascript
console.log(query.parse("/data=foo"));  //{ '/data': 'foo' }
console.log(query.parse("*data=foo"));  //{ '*data': 'foo' }
```
No problems. But with the ampersand case, you'll get:
```javascript
console.log(query.parse("&data=foo")); //{ '': null, data: 'foo' }
```
And i decided to make "clear" response {data: 'foo'} without {'': null}.

sindresorhus:
Alright, I guess it wouldn't hurt. Just fix the inline comments and I'll merge.

saromanov:
Fixed!

index.js
```diff
@@ -5,6 +5,10 @@ exports.parse = function (str) {
 		return {};
 	}
 
+	if (str[0] == '&') {
```
arthurvr: triple equal

index.js
sindresorhus: @saromanov :arrow_up:

saromanov:
Thanks, @arthurvr! Fixed, again.

sindresorhus:
@saromanov :arrow_right: https://github.com/sindresorhus/query-string/pull/23#discussion_r30956359

saromanov:
@sindresorhus , sorry, i not have much experience with github pull requests and i don't know what to do next. I was fixed with your comments(move to the right lines), but should i open new pull request? Or commit it again?

sindresorhus:
Sorry if I weren't clear. The `&` should be stripped here https://github.com/sindresorhus/query-string/pull/23/files#diff-168726dbe96b3ce427e7fedce31bb0bcR12 as we already strip `#` and `?` there.

saromanov:
Now, its clear for me :)

sindresorhus:
Looks good. Thanks :)

index.js
```diff
@@ -5,11 +5,7 @@ exports.parse = function (str) {
 		return {};
 	}
 
-	if (str[0] === '&') {
-		str = str.substr(1, str.length);
-	}
-
-	str = str.trim().replace(/^(\?|#)/, '');
+	str = str.trim().replace(/^(\?|#|&)/, '');
```
tmedetbekov: It is simpler and faster to use a regex `/^[?#&]/` instead

sindresorhus: I doubt it has any practical performance difference, but it is indeed simpler. Thanks :)

https://github.com/sindresorhus/query-string/commit/f2833411bbe975d38b261ed46c96edde4bd0e6e7

tmedetbekov: No big difference, but 7%

screen shot 2017-07-24 at 6 02 23 pm

Issue #15, opened 2015-01-26
add method to get query string from url


WickyNilliams:
For convenience, it would be nice to have a method that can pluck the query string from a url:

```js
var qs = require("query-string");
var url = "http://some.url/with?a=few&query=params";
var query = qs.getQuery(url); // "?a=few&query=params"
qs.parse(query); // { a : "few", query : "params" }
```

WickyNilliams:
perhaps a little naive, but implementation could be as simple as this?

```js
function getQuery(url) {
   return (url || "").split(/[\?\#]/)[1];
}
```

sindresorhus:
Why? Use `url.parse` in Node or `location.search` in the browser.

And yeah that implementation is too naive. Look at something like http://medialize.github.io/URI.js/ for how to do it correctly.

WickyNilliams:
I guess i was just a little surprised that it wouldn't query parse when given the full URL, and doesn't have a way to extract query string. I'm working in the browser, not parsing the current URL (so location.search isn't useful) but strings containing URLs.

Their implementation seems just as naive tbh, split on "?" https://github.com/medialize/URI.js/blob/891b5551da274d1ef3ff23ad3fdabea2b98344fd/src/URI.js#L406-L411

WickyNilliams:
I suppose I could:

```js
function getQuery(url) {
  var anchor = document.createElement("a");
  anchor.href = url;
  return anchor.search;
}
```

but that's a bit nasty imo

sindresorhus:
Not sure if it matters, but URI.js removes the fragment (`#`) before searching for the querystring `?`.

PR welcome :)

kud:
+1 for this.

I receive an url in a json content, and I had to do `response.url.split('?')[1]`.

adam-lynch:
+1

skibz:
+1

sindresorhus:
The next `+1` automatically volunteers to do a pull request ;)

kud:
+2

sindresorhus:
@kud I'll be expecting a pull request shortly.

kud:
rofl ;)

To be honest, I could work on it because I'm interested. I note it my (never) todo list.

skibz:
with pleasure

WickyNilliams:
@sindresorhus happy to take a look at this, being the person that created the issue and thus the plus-1-storm :D 

[...]

PR #24, opened 2015-06-01
ensure correct parsing the query params with '='

jhpoelen:
Expected:
http://someurl/?param=http%3A%2F%2Fsomeurl%3Fid%3D2837
should be parsed to:
{ param: 'http://someurl?id=2837' }
actual:
{ param: 'http://someurl?id' }

This change fixes that issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ exports.parse = function (str) {
 	return str.trim().split('&').reduce(function (ret, param) {
 		var parts = param.replace(/\+/g, ' ').split('=');
 		var key = parts[0];
-		var val = parts[1];
+		var val = parts.slice(1).join('=');
 
 		key = decodeURIComponent(key);
 		// missing `=` should be `null`:


sindresorhus:
No idea what your "fix" is doing, but it's not correct.

The reason it doesn't work is that supplying full URLs isn't supported yet: #15

jhpoelen:
@sindresorhus  Thanks for your quick response. I believe the issue is not related to the full url parsing: my example was a bit confusing... I would expect this test to pass:
```json
it('query strings params including embedded `=`', function () {
	assert.deepEqual(qs.parse('#param=http%3A%2F%2Fsomeurl%3Fid%3D2837'), {param: 'http://someurl?id=2837'});
});
```
The actual result is ```{ 'param': 'http://someurl?id' }```. 

I hope this clarifies the suggested fix.

sindresorhus:
You test seems to pass fine: https://travis-ci.org/sindresorhus/query-string/builds/64986110

Are you sure this is the query string lib you're using and you're using the latest version?

jhpoelen:
@sindresorhus thanks for responding and adding the test. The issue was on my end - I wasn't encoding the url properly in the href attribute. I owe you a beer ; )

PR #25, opened 2015-06-02
add getQueryString method fixes #15

### LINKED ISSUE #15: add method to get query string from url

WickyNilliams:
Adds a convenience method `getQueryString` to extract query string from a full URL

### DIFF
--- a/query-string.js
+++ b/query-string.js
@@ -7,7 +7,19 @@
 */
 (function () {
 	'use strict';
+
 	var queryString = {};
+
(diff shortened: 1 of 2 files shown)


WickyNilliams:
argh, should have pulled latest first. going to delete and try again

PR #26, opened 2015-06-02
add `.extract()` convenience method - fixes #15

### LINKED ISSUE #15: add method to get query string from url

WickyNilliams:
Add convenience method `getQueryString(url)` and assicoated tests

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-exports.getQueryString = function (maybeUrl) {
+exports.extract = function (maybeUrl) {
 	if (!maybeUrl) {
 		return;
 	}
--- a/index.js
+++ b/index.js
@@ -1,14 +1,14 @@
 'use strict';
 
 exports.getQueryString = function (maybeUrl) {
-  if(!maybeUrl) {
-    return;
-  }
+	if (!maybeUrl) {
+		return;
+	}
 
-  var pos = maybeUrl.indexOf('?');
-  if(pos > -1) {
-    return maybeUrl.substring(pos + 1);
-  }
+	var pos = maybeUrl.indexOf('?');
+	if (pos !== -1) {
+		return maybeUrl.substring(pos + 1);
+	}
 };
 
 exports.parse = function (str) {
--- a/index.js
+++ b/index.js
@@ -1,5 +1,16 @@
 'use strict';
 
+exports.getQueryString = function (maybeUrl) {
+  if(!maybeUrl) {
+    return;
+  }
+
+  var pos = maybeUrl.indexOf('?');
+  if(pos > -1) {
+    return maybeUrl.substring(pos + 1);
+  }
+};
+
 exports.parse = function (str) {
 	if (typeof str !== 'string') {
 		return {};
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1,5 +1,16 @@
 'use strict';
 
+exports.getQueryString = function (maybeUrl) {
+  if(!maybeUrl) {
+    return;
+  }
+
+  var pos = maybeUrl.indexOf('?');
+  if(pos > -1) {
```
kevva: `if (pos !== -1)`

index.js
```diff
@@ -1,5 +1,16 @@
 'use strict';
 
+exports.getQueryString = function (maybeUrl) {
+  if(!maybeUrl) {
```
kevva: `if (!maybeUrl)`

WickyNilliams:
Argh, sorry i even mixed spaces and tabs. I blame sublime for this. Fixed

index.js
```diff
@@ -1,5 +1,16 @@
 'use strict';
 
+exports.getQueryString = function (maybeUrl) {
```
sindresorhus: What do you think about it being `exports.extract`?

WickyNilliams: Your call really, I'm not precious about it. But personally I always prefer the most explicit thing possible, hence `getQueryString`

index.js
```diff
@@ -1,5 +1,16 @@
...
+	if (!maybeUrl) {
+		return;
+	}
+
+	var pos = maybeUrl.indexOf('?');
+	if (pos !== -1) {
+		return maybeUrl.substring(pos + 1);
+	}
+};
```
sindresorhus: How about just:

```js
exports.extract = function (str) {
	return str && str.split('?')[1];
};
```

index.js
sindresorhus: I too, but you have to read it as it's going to be used `queryString.extract()`. `queryString.getQueryString()` feels very verbose.

kevva: +1 for `.extract()`

index.js
WickyNilliams: Happy to make this change also. But be aware it means output is inconsistent for invalid input -empty string would return empty string instead of null.

index.js
WickyNilliams: done

index.js
WickyNilliams: Maybe that's OK though. Let me know and I'll make changes if necessary

sindresorhus: I think that's actually beneficial. It should return a string as long as the input is a string. And it should throw if the input is not a string, so:

```js
exports.extract = function (str) {
    return str.split('?')[1] || '';
};
```

WickyNilliams: OK done that, and updated tests to match what you described above.

WickyNilliams:
👍

PR #29, opened 2015-07-21
use more stringent version of encodeURIComponent

go-oleg:


kevva:
https://github.com/kevva/strict-uri-encode

sindresorhus:
Why is this needed though? What makes the builtin method not good enough?

go-oleg:
My specific issue is that [Confluence](https://www.atlassian.com/software/confluence) will truncate the "link" of a pasted URL at the first `'`...so they must be following stricter rules w.r.t. what is and isn't a valid URL character. At first I thought it was a Confluence bug, but seeing the snippet on MDN it makes more sense. Perhaps there are other products that share this behavior out there.

sindresorhus:
Ok, sure. I guess it doesn't hurt to be stricter. Will merge when you've changed it to use https://github.com/kevva/strict-uri-encode ;)

go-oleg:
@sindresorhus, done ;-).

kevva:
Thanks yo :).

jsg2021:
Now stringify throws for non-string values...

```
{
    page: 0, 
    size: 5
}
```

This makes this utility package less convenient.

jsg2021:
the strict encode is nice, but its type check makes auto-type coercion for the base simple case a little more painful.

kevva:
@jsg2021, fixed in https://github.com/kevva/strict-uri-encode/commit/5d1d747e5fb2ad191d72eda3be146e554accfc2e.

PR #36, opened 2015-10-01
Undefined values are removed from stringified output

tlhunter:
When stringifying an object where a value is undefined, previously the value would be passed along as `'undefined'`. Now we remove it from the resulting query string entirely. This makes `qs.stringify()` more consistent with `JSON.stringify()`.


sindresorhus:
This is inconsistent with the Node.js `querystring` module though.

```
❯ node
> querystring.stringify({abc: undefined, foo: 'baz'})
'abc=&foo=baz'
```

Not really sure which one is correct. What do other query string modules do? Like https://github.com/hapijs/qs

tlhunter:
Hapi's qs library does omit undefined:

https://github.com/hapijs/qs/blob/master/test/stringify.js#L169

SamVerschueren:
I guess @tlhunter is right. I dived into the code (didn't tested it) but it seems that `qs.parse('?val=undefined');` will result in `{val: 'undefined'}`. Couldn't find a test for this so not 100% sure though.

So option 1 is omit the property, option 2 is letting stringify only omit the value but not the key. `qs.stringify({foo: undefined, bar: 1});` will result in `?foo&bar=1`.

SamVerschueren:
I checked it with the following small example

```js
'use strict';

var qs = require('query-string');

console.log(qs.parse(qs.stringify({foo: undefined, bar: 1})));
```

The output is
```
{ bar: '1', foo: 'undefined' }
```

Which is not entirely correct off course.

sindresorhus:
The questions is, is there any value in the result of stringify then parse being identical to the input?

```js
test(t => {
	var x = {foo: undefined};
	t.same(fn.parse(fn.stringify(x)), x);
	t.end();
});
```

sindresorhus:
Maybe `undefined` should be missing key & value and `null` should be missing value? That would be compatible with this check: https://github.com/sindresorhus/query-string/blob/8c6908ed370963cb76f2d9f874b8f370dbc27987/index.js#L28

SamVerschueren:
Seems good!

SamVerschueren:
@tlhunter Care to give it a go? The implementation should be

```js
if (val === undefined) {
	return '';
}
if (val === null) {
	return key;
}
```

+ extra tests

tlhunter:
I'll give it a go

sindresorhus:
@tlhunter Can you fix the merge conflict?

tlhunter:
@sindresorhus fixed

PR #38, opened 2015-10-04
Split query params using a regex

steveluscher:
Saves an array (`shift`) and a string (`join`) manipulation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,11 +17,11 @@ exports.parse = function (str) {
 	}
 
 	return str.split('&').reduce(function (ret, param) {
-		var parts = param.replace(/\+/g, ' ').split('=');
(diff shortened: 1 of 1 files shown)


steveluscher:
One moment please.

steveluscher:
That should do it.

Note: This will probably make this library fail on IE8, if you care.

sindresorhus:
> This will probably make this library fail on IE8, if you care.

I don't, but removing support now would be a breaking change and for little gain. Why would it not work in IE8?

steveluscher:
Just fired up IE8 to try it out:

screen shot 2015-10-10 at 1 09 06 am

SamVerschueren:
What about splitting with indexOf?

steveluscher:
Yep! Could do that.

sindresorhus:
ping @steveluscher

PR #41, opened 2015-12-03
Make tests compatible with latest version of AVA

bpugh:
Tests were throwing the 't.end is not supported in this context.' with
latest version of AVA so I removed all the t.end() calls since it seems
to no longer be necessary. May also want to consider specifying a version for AVA
in package.json.


sindresorhus:
> May also want to consider specifying a version for AVA
in package.json.

Nah, this was probably the last breaking change ever. Don't want to have to constantly update AVA of hundreds of tiny modules.

bpugh:
Makes sense. Thanks for all the handy modules!

PR #45, opened 2016-02-26
Update readme.md

jaimeagudo:
Leading `?` is not generated and confuses users


sindresorhus:
It is. Go to https://github.com/sindresorhus/query-string/pull/45/files?w=1 and in the console:

```
location.search
"?w=1"
```

PR #48, opened 2016-03-17
support object properties

### LINKED ISSUE #47: `require('query-string').parse('hasOwnProperty&hasOwnProperty');` breaks
Unsafe usage of `hasOwnProperty` means that parsing certain query strings will throw an error.

The following cases are also not working correctly:

    qs.parse('__proto__');
    qs.parse('__proto__&__proto__');

`Object.create(null)` should fix the problem.

SamVerschueren:
I'm not 100% sure if this is the correct solution for #47, but just gave it a go.

The tests fail although every result is correct. Because the prototype of the object is set to null, `t.same` does not validate correctly. Not sure how to resolve this.

-

Fixes #47.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,7 +29,7 @@ exports.parse = function (str) {
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
 		val = val === undefined ? null : decodeURIComponent(val);
 
-		if (!ret.hasOwnProperty(key)) {
+		if (!ret[key]) {
 			ret[key] = val;
 		} else if (Array.isArray(ret[key])) {
 			ret[key].push(val);
@@ -38,7 +38,7 @@ exports.parse = function (str) {
 		}
 
 		return ret;
-	}, {});
+	}, Object.create(null));
 };
 
 exports.stringify = function (obj) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Because the prototype of the object is set to null, t.same does not validate correctly. Not sure how to resolve this.

Was fixed here: https://github.com/sindresorhus/ava/commit/a7826cf1c81a38eaa0908c8ae37525e82a43c192 Try master.

For now I guess you could just check the property directly with `t.is` instead.

sindresorhus:
LGTM for the code.

SamVerschueren:
Added a helper method so that we could still use the same syntax als `t.same` and it will be easier to revert back in the future once the next AVA is released.

sindresorhus:
Thanks @SamVerschueren. You're awesome :)

SamVerschueren:
You're welcome :)!

index.js
```diff
@@ -29,7 +29,7 @@ exports.parse = function (str) {
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
 		val = val === undefined ? null : decodeURIComponent(val);
 
-		if (!ret.hasOwnProperty(key)) {
+		if (!ret[key]) {
```
cspotcode: Thank you for fixing this!  However, I think this conditional is incorrect when the value is null or the empty string.
For example `require('query-string').parse('a&a=&a=2')`

PR #51, opened 2016-03-20
handle `null` and `undefined` in array - fixes #49

### LINKED ISSUE #49: in arrays passed to stringify, `null` the same as string `"null"`
I'm not sure if this is a bug; maybe it's intended or unspecified behavior.  Within an array, `null` is the same as the string `"null"`.

    var params = {'a': null, 'b': [null, 123]};
    require('querystring').stringify(params);
    // Node core: "a=&b=&b=123"
    require('query-string').stringify(params);
    // query-string: "a&b=123&b=null"

Here's my rationale:

    var qs = require('query-string');
    // If these are equivalent:
    qs.stringify({a: 123}) === qs.stringify({a: [123]});
    // These should be as well?
    qs.stringify({a: null}) === qs.stringify({a: [null]});

SamVerschueren:
This PR fixes #49.

@sindresorhus I know you are not a big fan of `Array.prototype.reduce`, so I used a simple `forEach` instead (which actually looks cleaner as well).

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,9 +54,21 @@ exports.stringify = function (obj) {
 		}
 
 		if (Array.isArray(val)) {
-			return val.slice().sort().map(function (val2) {
-				return strictUriEncode(key) + '=' + strictUriEncode(val2);
-			}).join('&');
+			var result = [];
+
+			val.slice().sort().forEach(function (val2) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
😎👍

PR #52, opened 2016-03-24
Add 'lowercaseHex' option to stringify()

moroshko:
Sometimes, when building URLs, you need to be able to control the casing of the query string, including the URL encoded hex codes (e.g. `%2f` vs `%2F`). This is important for SEO reasons.

By default, `stringify()` returns uppercase hex codes. This PR adds a `lowercaseHex` option to lowercase them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,8 +41,12 @@ exports.parse = function (str) {
 	}, {});
 };
 
-exports.stringify = function (obj) {
-	return obj ? Object.keys(obj).sort().map(function (key) {
+var hexRegex = /%[0-9A-F]{2}/g;
+
+exports.stringify = function (obj, options) {
+	options = options || {};
+
+	var queryString = obj ? Object.keys(obj).sort().map(function (key) {
 		var val = obj[key];
 
 		if (val === undefined) {
@@ -75,4 +79,12 @@ exports.stringify = function (obj) {
 	}).filter(function (x) {
 		return x.length > 0;
 	}).join('&') : '';
+
+	if (options.lowercaseHex === true) {
+		return queryString.replace(hexRegex, function (hex) {
+			return hex.toLowerCase();
+		});
+	}
+
+	return queryString;
 };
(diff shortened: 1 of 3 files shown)


sindresorhus:
> This is important for SEO reasons.

Source?

moroshko:
It's hard to find an official source for this, but here are some pointers:

> URLs in general are case-sensitive (with the exception of machine names). There may be URLs, or parts of URLs, where case doesn't matter, but identifying these may not be easy. Users should always consider that URLs are case-sensitive.

[Source](https://www.w3.org/TR/WD-html40-970708/htmlweb.html#htmlurl)

> Mixed case URLs can be a source of duplicate content.

[Source](http://searchengineland.com/seo-friendly-url-syntax-practices-134218)

> Hi John, based on the existing standards, URLs are case-sensitive, so yes, these would be seen as separate URLs. Since the content on the URLs is the same, we'll generally recognize that and only keep one of them. However, we'd recommend that you try to keep all links going to one version of the URL.

[Source](http://webmasters.stackexchange.com/a/69719/11516)

See also [this one](https://www.webmasterworld.com/webmaster/3929874.htm).

moroshko:
@sindresorhus Note that I updated the `stringify()` API to have `options` to make it more expressive:

`stringify({ foo: 'bar' }, true)` - Not clear what `true` means
`stringify({ foo: 'bar' }, { lowercaseHex: true })` - No magical true :)

sindresorhus:
Thanks for elaborating, but I'm unfortunately going to pass on this. I've never needed or seen it being needed at any time and I've been using query strings for many years. None of the other query string modules I've seen has such option either. I just don't think it's worth an option here. Sorry.

Also:

> If two URIs differ only in the case of hexadecimal digits used in percent-encoded octets, they are equivalent. - http://stackoverflow.com/a/18347870/64949

-

I would however be happy to accept a PR that adds normalization of the escape to https://github.com/sindresorhus/normalize-url with an option for lower or upper case ;)

-

You could also try getting it added to https://github.com/ljharb/qs.

Issue #50, opened 2016-03-18
result from `parse` no longer has `hasOwnProperty` method


SpainTrain:
in 3.0.1, `parse` had the following behavior:

```
> qs.parse('my-param=my-val')
{ 'my-param': 'my-val' }
> qs.parse('my-param=my-val').hasOwnProperty
[Function: hasOwnProperty]
> qs.parse('h=j').hasOwnProperty
[Function: hasOwnProperty]
> qs.parse('').hasOwnProperty
[Function: hasOwnProperty]
```

in 3.0.2, the object returned no longer has the `hasOwnProperty` method:

```
> qs.parse('hbud-fixture=not-logged-in')
{ 'hbud-fixture': 'not-logged-in' }
> qs.parse('hbud-fixture=not-logged-in').hasOwnProperty
undefined
> qs.parse('hbudfixture=not-logged-in').hasOwnProperty
undefined
> qs.parse('hbudfixture=notloggedin').hasOwnProperty
undefined
> qs.parse('').hasOwnProperty
[Function: hasOwnProperty]
> qs.parse('h=j').hasOwnProperty
undefined
```

I am unsure whether this was considered part of the API, but some consumers rely on the behavior - https://github.com/mjackson/history/blob/master/modules/useQueries.js#L13

SamVerschueren:
This is because of https://github.com/sindresorhus/query-string/issues/47

`hasOwnProperty` could be a query string property which interfered with the `hasOwnProperty` method of the object. Maybe the change should've been released as breaking @sindresorhus ?

taion:
:+1: on this – #47 is a good change, but it'd be best to roll it back and release a v3.0.3 to avoid breaking people who are using the `for ... in`/`hasOwnProperty` pattern, then re-release as v4.0.0 so we can update our code.

taion:
Actually, on second thought, this is somewhat problematic – it means that we need to not use `obj.hasOwnProperty` on anything coming out of `parseQuery` here, which is quite a lot to keep track of.

cspotcode:
@taion: The `for ... in` / `hasOwnProperty` pattern should usually be done like this:

```
for(var key in obj) {
    if(Object.prototype.hasOwnProperty.call(obj, key) {
        // ...
    }
```

However, to avoid using `Object.create(null)` in `3.x.x` you could use `Object.defineProperty()` when attaching properties to the returned object.  That would avoid issues with `__proto__`.

[...]

PR #53, opened 2016-03-24
support object properties - fixes #47

### LINKED ISSUE #47: `require('query-string').parse('hasOwnProperty&hasOwnProperty');` breaks

SamVerschueren:
Fixes #47 and ends the discussion in #50. Should be released as breaking change (e.g. `4.x`). Added some improvements over the previous PR regarding this issue and added a note in the documentation. Feel free to provide any feedback :).

// @cspotcode

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,17 +6,21 @@ exports.extract = function (str) {
 };
 
 exports.parse = function (str) {
+	// Create an object with no prototype
+	// https://github.com/sindresorhus/query-string/issues/47
+	var ret = Object.create(null);
+
 	if (typeof str !== 'string') {
-		return {};
+		return ret;
 	}
 
 	str = str.trim().replace(/^(\?|#|&)/, '');
 
 	if (!str) {
-		return {};
+		return ret;
 	}
 
-	return str.split('&').reduce(function (ret, param) {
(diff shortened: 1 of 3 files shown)


cspotcode:
Thanks!  There's a small regression with these two test cases:

```
queryString.parse('a&a'); // should be {a: [null, null]} but returns {a: null}
queryString.parse('a=&a'); // should be {a: ['', null] but returns {a: null}
```

This was caused by replacing the `hasOwnProperty` check with a truthiness check.  Both `null` and the empty string are falsy values.

SamVerschueren:
Good catch! Will fix asap!

SamVerschueren:
Should be fixed.

cspotcode:
Sweet, looks good to me.

SamVerschueren:
Thanks Sindre :+1:!

sindresorhus:
Thanks for the PR :)

PR #54, opened 2016-03-26
add strict option

### LINKED ISSUE #42: Is strict-uri-encode really necessary?
I am using https://github.com/Sage/jsurl to efficiently encode JSON objects in the search string, and it uses the looser URI encoding range used by browsers, including the `'` character.

I also use rackt/history which uses query-string, which converts `'` to `%27`, which grows my urls, which makes me sad. Is there a reason the strict URI range should be used?

SamVerschueren:
This PR should resolve #42.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,10 @@
 'use strict';
 var strictUriEncode = require('strict-uri-encode');
 
+function encode(value, strict) {
+	return strict ? strictUriEncode(value) : encodeURIComponent(value);
+}
+
 exports.extract = function (str) {
 	return str.split('?')[1] || '';
 };
@@ -45,7 +49,11 @@ exports.parse = function (str) {
 	return ret;
 };
 
-exports.stringify = function (obj) {
+exports.stringify = function (obj, opts) {
+	opts = opts || {};
+
+	var strict = opts.strict !== false;
+
 	return obj ? Object.keys(obj).sort().map(function (key) {
 		var val = obj[key];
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -45,7 +49,10 @@ exports.parse = function (str) {
 	return ret;
 };
 
-exports.stringify = function (obj) {
+exports.stringify = function (obj, opts) {
+	opts = opts || {};
+	opts.strict = opts.strict !== false;
```
sindresorhus: Don't overwrite the user supplied options object. Objects are reference based, so this will change the behavior if the user calls multiple time with the same object. Instead just assign it to a variable.

readme.md
```diff
@@ -47,10 +47,18 @@ Parse a query string into an object. Leading `?` or `#` are ignored, so you can
...
 
 Stringify an object into a query string, sorting the keys.
 
+#### strict
+
+Type: `boolean`
+Default: `true`
+
+Strictly encode URI components with [strict-uri-encode](https://github.com/kevva/strict-uri-encode). It uses [encodeURIComponent](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
+if set to false.
```
sindresorhus: Maybe mention something like "You probably don't care about this" and link to the issue thread.

SamVerschueren: Something like

> Strictly encode URI components with [strict-uri-encode](https://github.com/kevva/strict-uri-encode). It uses [encodeURIComponent](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
if set to false. You probably don't care about [this](https://github.com/sindresorhus/query-string/issues/42).

or

> Strictly encode URI components with [strict-uri-encode](https://github.com/kevva/strict-uri-encode). It uses [encodeURIComponent](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
if set to false. You probably don't care about this. See [#42](https://github.com/sindresorhus/query-string/issues/42) for more information.

sindresorhus: > Strictly encode URI components with [strict-uri-encode](https://github.com/kevva/strict-uri-encode). It uses [encodeURIComponent](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)
if set to false. You probably [don't care](https://github.com/sindresorhus/query-string/issues/42) about this option.

SamVerschueren:
done

sindresorhus:
🎉👍

SamVerschueren:
Awesome!

seanmunson:
... So something like :

`querystring.stringify(argObject,,{strict:false}) `

should work ?

PR #55, opened 2016-03-26
Fix undefined check and tests

akhaku:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "query-string",
-  "version": "4.0.0",
+  "version": "4.0.1",
```
SamVerschueren: Do not bump the version

index.js
```diff
@@ -33,7 +33,7 @@ exports.parse = function (str) {
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
 		val = val === undefined ? null : decodeURIComponent(val);
 
-		if (ret[key] !== undefined) {
+		if (ret[key] === undefined) {
```
SamVerschueren: Why wasn't this catched by the tests? :o

test/parse.js
```diff
@@ -4,67 +4,67 @@ import fn from '../';
 // https://github.com/sindresorhus/query-string/pull/48#issuecomment-198242935
 function tsame(t, actual, expected) {
 	Object.keys(expected).forEach(key => {
-		t.is(actual[key], expected[key]);
+		if (Array.isArray(expected[key])) {
+			tsame(t, actual[key], expected[key]);
```
SamVerschueren:
```js
t.same(actual[key], expected[key]);
```

test/parse.js
```diff
@@ -4,67 +4,67 @@ import fn from '../';
...
+		} else {
+			t.is(actual[key], expected[key]);
+		}
 	});
 }
 
-test.beforeEach(t => {
-	t.context.same = tsame.bind(undefined, t);
-});
```
SamVerschueren: Why remove it?

test/parse.js
```diff
@@ -4,67 +4,67 @@ import fn from '../';
...
 test('query strings params including raw `=`', t => {
-	t.context.same(fn.parse('?param=http://someurl?id=2837'), {param: 'http://someurl?id=2837'});
+	tsame(t, fn.parse('?param=http://someurl?id=2837'), {param: 'http://someurl?id=2837'});
 });
 
 test('object properties', t => {
 	t.notOk(fn.parse().prototype);
-	t.context.same(fn.parse('hasOwnProperty=foo'), {hasOwnProperty: 'fo'});
-	t.context.same(fn.parse('__proto__=bar'), {__proto__: 'bar'});
+	tsame(t, fn.parse('hasOwnProperty=foo'), {hasOwnProperty: 'foo'});
```
SamVerschueren: Took me a while to figure out why this didn't come up in the tests. I think the expected object has to be instantiated with `Object.create(null)`.

package.json
akhaku: sure, will revert

index.js
akhaku: See the commit message for the commit fixing the tests, but turns out refactoring to bind the tsame function broke ava's assertions somehow. Don't completely understand - stepped through and saw it hitting the `throw new assert.AssertionError ` line in ava/lib/assert.js, but that doesn't do anything - my javascript fu isn't strong enough.

test/parse.js
akhaku: sounds good

test/parse.js
akhaku: Not being used anymore - I'm explicitly passing in `t` to the tsame function. See comment above (and commit message for 6d80fe9469) about binding the function to the the test instance causing behaviour I don't fully understand.

test/parse.js
akhaku: Possibly, but it may have more to do with the tests being completely broken :) see comments above for more detail.

SamVerschueren:
Thanks! Will have more detailed look later today. Want to clear some things up why it didn't work out in the first place :)

SamVerschueren:
Alright, all the tests failed because of the `tsame.bind(undefined, t)` in the `beforeEach`. The `t` in the `beforeEach` isn't the same `t` as in the test itself. So when the `beforeEach` test (because it's also a test on it's own) has ended, it can't assert other values anymore off course. So that's why the validations didn't come up.

test/parse.js
SamVerschueren: Looks like it isn't necessary though.

SamVerschueren:
@sindresorhus LGTM

sindresorhus:
:+1: Thanks for fixing.

akhaku:
Thanks for the quick turnaround!

PR #57, opened 2016-04-12
:arrow_up: ava@0.14.0

SamVerschueren:
bumped AVA

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -1,70 +1,58 @@
 import test from 'ava';
 import fn from '../';
 
-// https://github.com/sindresorhus/query-string/pull/48#issuecomment-198242935
-function tsame(t, actual, expected) {
-	Object.keys(expected).forEach(key => {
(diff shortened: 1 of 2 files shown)


SamVerschueren:
Apparently `Object.create(null).__proto__` returns `null` on node 0.10 and `undefined` on node 0.12+.

sindresorhus:
Weird, just use `t.true(Object.create(null).__proto__ == null)` with a comment then.

SamVerschueren:
It's this test that's failing

```js
t.is(fn.parse('__proto__=bar').__proto__, 'bar');
```

This returns `[null, 'bar']` on node 0.10 which is really an edge case. So I can drop the test if you want.

sindresorhus:
Yeah, drop it.

SamVerschueren:
Done. Don't think it needs other tweaks?

sindresorhus:
:+1: :fireworks:

PR #60, opened 2016-05-05
Add option to stringify without encoding

Pholey:


sindresorhus:
Why?

mehcode:
@sindresorhus We have an application that talks to an API. The API expects raw colons `:`in the query string values. We can't change the API. It'd be nice to construct the query from object form.

Pholey:
Not to mention being able to generate a human readable query string is great for debugging

index.js
```diff
@@ -52,7 +56,14 @@ exports.parse = function (str) {
 exports.stringify = function (obj, opts) {
 	opts = opts || {};
 
-	var strict = opts.strict !== false;
+	// Default encode to true to maintain previous behavior
+	if (opts.encode === undefined) {
+		opts.encode = true;
```
sindresorhus: Here you're overwriting the user-supplied object, which means if the user calls this method twice with the same object it will have different behavior. Remember that objects are reference based.

SamVerschueren: Use [object-assign](https://github.com/sindresorhus/object-assign) to solve it.

sindresorhus:
Alright. Can you also add it to the docs?

PR #62, opened 2016-05-31
document support for foo=bar&foo=baz

moredip:
add info about supporting multiple instances of the same key


sindresorhus:
👍 Thanks :)

PR #65, opened 2016-06-09
handle possible error of decodeURIComponent method

pd4d10:
Hi,

JS builtin method `decodeURIComponent` will throw an error if a value could not be decoded.
```
> decodeURIComponent('param=%')
Uncaught URIError: URI malformed(…)
```
In this PR a function `decode` was added to handle this error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,14 @@ function encode(value, opts) {
 	return value;
 }
 
+function decode(value) {
+	try {
+		return decodeURIComponent(value);
+	} catch (err) {
+		return value;
+	}
+}
+
 exports.extract = function (str) {
 	return str.split('?')[1] || '';
 };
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -36,11 +44,11 @@ exports.parse = function (str) {
 		var key = parts.shift();
 		var val = parts.length > 0 ? parts.join('=') : undefined;
 
-		key = decodeURIComponent(key);
+		key = decode(key);
 
 		// missing `=` should be `null`:
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
-		val = val === undefined ? null : decodeURIComponent(val);
+		val = val === undefined ? null : decode(val);
```
SamVerschueren: @sindresorhus should we move this check to the `decode` function?

sindresorhus: Why?

sindresorhus:
I don't think this is a good idea. `%` is not a valid query param value. You need to encode it to `%25` first.

See: http://stackoverflow.com/questions/7449588/why-does-decodeuricomponent-lock-up-my-browser

If we handle this, what if the value is `%%23`. Then we don't decode the `%23`.

pd4d10:
OK. Actually I find this problem when using `react-router`, while `react-router` depends on `history` depends on `query-string`.

I've opened an issue on `history` to discuss this, see:
https://github.com/ReactJSTraining/history/issues/312

cthorner:
Just ran into this as well and I'm handling it by failing gracefully with a try catch statement so that my entire site doesn't blow up.

PR #67, opened 2016-06-10
Encode key when value is null

### LINKED ISSUE #64: Why is key not encoded when value is null
https://github.com/sindresorhus/query-string/blob/master/index.js#L73

Why is it not being encoded? A small comment might help...

Aweary:
Resolves #64 

~~I had to add a linting exception for `quote-props` since the key needs to be quoted and the value needs to be `null` for the tests to work. If you have another preferred way to handle that, let me know~~

### DIFF
--- a/test/stringify.js
+++ b/test/stringify.js
@@ -2,62 +2,62 @@ import test from 'ava';
 import fn from '../';
 
 test('stringify', t => {
-	t.deepEqual(fn.stringify({foo: 'bar'}), 'foo=bar');
-	t.deepEqual(fn.stringify({foo: 'bar', bar: 'baz'}), 'bar=baz&foo=bar');
+	t.is(fn.stringify({foo: 'bar'}), 'foo=bar');
+	t.is(fn.stringify({foo: 'bar', bar: 'baz'}), 'bar=baz&foo=bar');
 });
 
 test('different types', t => {
-	t.deepEqual(fn.stringify(), '');
-	t.deepEqual(fn.stringify(0), '');
+	t.is(fn.stringify(), '');
+	t.is(fn.stringify(0), '');
 });
 
 test('URI encode', t => {
-	t.deepEqual(fn.stringify({'foo bar': 'baz faz'}), 'foo%20bar=baz%20faz');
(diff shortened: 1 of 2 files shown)


test/stringify.js
```diff
@@ -37,7 +37,8 @@ test('should not encode undefined values', t => {
 });
 
 test('should encode null values as just a key', t => {
-	t.deepEqual(fn.stringify({xyz: null, abc: null, foo: 'baz'}), 'abc&foo=baz&xyz');
+	// eslint-disable-next-line quote-props
+	t.deepEqual(fn.stringify({'x y z': null, abc: null, foo: 'baz'}), 'abc&foo=baz&x%20y%20z');
```
SamVerschueren: Doesn't it work if you quote them all instead of only that single one? Not sure though, just guessing :).

Aweary: Good guess, you were right 😄  fixed and squashed.

test/stringify.js
```diff
@@ -37,7 +37,7 @@ test('should not encode undefined values', t => {
 });
 
 test('should encode null values as just a key', t => {
-	t.deepEqual(fn.stringify({xyz: null, abc: null, foo: 'baz'}), 'abc&foo=baz&xyz');
+	t.deepEqual(fn.stringify({'x y z': null, 'abc': null, 'foo': 'baz'}), 'abc&foo=baz&x%20y%20z');
```
SamVerschueren: Use `t.is` here

Aweary: Where, in place of `deepEqual`?

Aweary: All I did was augment an existing test, btw. If you want to use `t.is` do you want me to change all the other tests too while I'm at it?

SamVerschueren: Yes it should be `t.is`, `deepEqual` is for objects.

Aweary: 👍 changed all the tests in `test/stringify.js` to use `t.is` instead

SamVerschueren:
Thanks @Aweary ! :beers:

Mouvedia:
#64 was "Why is key not encoded when value is null"

My question is what motivated that change?

PR #68, opened 2016-06-16
Update usage example. stringify does not include '?'

alyssaq:


SamVerschueren:
Actually, if you open your chrome devtools and take the following steps

1. `location.search="foo=bar";`
2. <kbd>Enter</kbd>
3. `location.search`
4. <kbd>Enter</kbd>

The output will be `?foo=bar`. So in my opinion, the example is correct.

sindresorhus:
I've improved the usage example: https://github.com/sindresorhus/query-string/commit/47788b675646b0fd14e4e4adae439a4b51fb71b2 Hope that's clearer :)

alyssaq:
Much clearer. Thanks!

Issue #66, opened 2016-06-10
Incorrect handling of arrays?


Anahkiasen:
Currently if `query-string` encounters an array it encodes it without brackets:

``` js
> require('query-string').stringify({foo: ['bar', 'baz']})
'foo=bar&foo=baz'
```

However that seems to go against the default browser behavior or one of other languages such as PHP. Shouldn't it be `foo[]=bar&foo[]=baz`?

I know this is kind of a grey area of the RFC but that struck me as odd, is this intended?

SamVerschueren:
I believe this is intended. All node frameworks handle it correctly.

abdulrahmancz:
What about adding config option for stringify and parse methods?

johannesnagl:
The Gozala/querystring repo has the same PR: https://github.com/Gozala/querystring/pull/10

sindresorhus:
Yeah sure, if someone does a good pull request with tests.

rapzo:
See #79

sindresorhus:
Fixed by #79.

PR #70, opened 2016-06-27
Incorrect handling of arrays

abdulrahmancz:
Add square brackets for array items of stringified query string. Remove square brackets while parsing query string with array items. Updated tests and readme. Also related ti [issue#66](https://github.com/sindresorhus/query-string/issues/66).

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,7 +36,8 @@ exports.parse = function (str) {
 		var key = parts.shift();
 		var val = parts.length > 0 ? parts.join('=') : undefined;
 
-		key = decodeURIComponent(key);
+		key = decodeURIComponent(key)
+			.replace(/\[\]$/,"");
 
 		// missing `=` should be `null`:
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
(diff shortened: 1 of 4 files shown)


Aweary:
This is inconsistent with how the native `querystring` module works in Node.

```js
> querystring.stringify({ foo: ['bar', 'baz'] })
'foo=bar&foo=baz'
```

If the official specification doesn't clarify this behavior then it seems most reasonable to remain consistent with the platform.

kevva:
This should definitely be behind a option like it is in https://github.com/ljharb/qs.

index.js
```diff
@@ -80,11 +81,12 @@ exports.stringify = function (obj, opts) {
 				if (val2 === undefined) {
 					return;
 				}
+				var arrayKey = encode(key, encode) + '[]';
```
kevva: Add an empty line before this.

index.js
```diff
@@ -36,7 +36,8 @@ exports.parse = function (str) {
 		var key = parts.shift();
 		var val = parts.length > 0 ? parts.join('=') : undefined;
 
-		key = decodeURIComponent(key);
+		key = decodeURIComponent(key)
+			.replace(/\[\]$/, '');
```
kevva: `.replace` doesn't need to be on its own line.

sindresorhus:
Closing for lack of response.

PR #71, opened 2016-07-11
Support ordering by allowing array of object to be passed to stringif…

ivnmaksimovic:
Support ordering by allowing array of object to be passed to stringify. Still using same stringify implementation. Test added and documentation updated.
Original stringify function is extracted to stringifyObject function. This way I was able to use it without any changes to it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -62,7 +62,19 @@ exports.stringify = function (obj, opts) {
 
 	opts = objectAssign(defaults, opts);
 
-	return obj ? Object.keys(obj).sort().map(function (key) {
+	if (!obj) {
+		return '';
+	} else if (Array.isArray(obj)) {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Why?

ivnmaksimovic:
1. nice: `category=unicorns&subcategory=horns&color=white&sort=length`
2. not nice: `category=unicorns&sort=length&color=white&subcategory=horns`

Case 2 might not look exactly as example...

sindresorhus:
I appreciate the suggestion, but this is not something I'm interested. I've already accepted more options than I'd like. I suggest forking and creating a more feature-rich module or maybe try getting it into https://github.com/ljharb/qs.

PR #75, opened 2016-10-03
Add a tip how to convert parse result to have default prototype

axelpale:
See this StackOverflow discussion: http://stackoverflow.com/q/39839991/638546

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -54,7 +54,7 @@ console.log(location.search);
 
 Parse a query string into an object. Leading `?` or `#` are ignored, so you can pass `location.search` or `location.hash` directly.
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
As pointed out by http://stackoverflow.com/a/39840363/64949, you shouldn't...

SamVerschueren:
And that should also be the accepted answer...

axelpale:
@sindresorhus Thanks for considering. However, I would like to emphasise the risk if the PR is not merged. The PR also points out that security issue. For security purposes, I think it is better to explicitly mention the issue in the readme (like done in the PR) than let users to figure it out themselves. As the insecure solution (Object.assign) is so simple, I am afraid some users implement that and never realise its drawbacks if not clearly mentioned in the readme. Thus, I ask you to reconsider.

PR #77, opened 2016-11-08
Added option `sortKeys` as a control to avoid sorting the keys

maurizzzio:
Defaults to `true`, every other test case remains intact

```js
import { stringify } from 'query-string';
stringify({
  xyz: 'xyz',
  abc: 'abc'
}, {sortKeys: false}))  // 'xyz=xyz&abc=abc' 
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,42 +57,49 @@ exports.parse = function (str) {
 exports.stringify = function (obj, opts) {
 	var defaults = {
 		encode: true,
+		sortKeys: true,
 		strict: true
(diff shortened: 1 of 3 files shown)


sindresorhus:
Why? What problem does this solve?

maurizzzio:
Sorry about that, I just saw https://github.com/sindresorhus/query-string/pull/71 and even though personally I'd let the sort to be optional you're right about keeping it simple, thanks for looking at this!

PR #78, opened 2016-11-10
Correct handling of arrays

rapzo:
Highlighted in issue #66


satazor:
As mentioned in #66, I think the change in this behaviour should be put behind a option. Additionally, the README should include documentation for this new option.

sindresorhus:
Look at the previous attempt for this and its review:
https://github.com/sindresorhus/query-string/pull/70

rapzo:
Bummer, didn't look at previous tries. My bad.
Gonna pick up from where #70 left.

PR #91, opened 2017-04-18
Prevent non-array values from being wrapped in arrays

glasnoster:
#89 (`4.3.3`) broke non-array values when `arrayFormat` was set to `bracket`.

`foo=bar` resulted in `{foo: ['bar']}` instead of `{foo: 'bar'}`


sindresorhus:
// @rapzo

rapzo:
Nice catch! It felt like there had to be some case the previous fix was breaking. Thanks for adding tests that cover it!

PR #92, opened 2017-04-18
Allow ';' as query string delimiter as well as &.

neomorphic:
There is conflicting information on what is or is not accepted as a
query delimiter string, but at some point in history, both & and ; have
been used.

https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax

Since the ';' is only allowed as a delimiter in the url, it seems that we
should be be able to parse the query string by splitting on the ; in the
same way we split on the &.

### DIFF
--- a/index.js
+++ b/index.js
@@ -133,7 +133,7 @@ exports.parse = function (str, opts) {
 		return ret;
 	}
 
-	str.split('&').forEach(function (param) {
+	str.split(/&|;/).forEach(function (param) {
 		var parts = param.replace(/\+/g, ' ').split('=');
 		// Firefox (pre 40) decodes `%3D` to `=`
 		// https://github.com/sindresorhus/query-string/pull/37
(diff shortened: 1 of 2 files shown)


SamVerschueren:
I don't see any good reason why we should do this. As pointed out by yourself and by the footnote

> Historic RFC 1866 (obsoleted by RFC 2854) encourages CGI authors to support ';' in addition to '&'

I never came across this in the wild (could be me though) and we shouldn't support it just because some servers in 1980 used this format. There's also a reason why we are dropping support for Node.js 0.10 and 0.12 and choose not to maintain deprecated stuff.

So feel free to come up with arguments why you think it should be added :).

neomorphic:
Not got a lot of spare cycles to argue for it, just looked like a quick change that wouldn’t break anything else, since ; is only allowed as a delimiter.
I will kill the pull request.
--
J.

PR #99, opened 2017-06-29
Add better component decoder - fixes #98

### LINKED ISSUE #98: URIError: malformed URI sequence
When you reload a page with a URL having the form '?q=%'  an error occurs **URIError: malformed URI sequence** 

Solution: 

`        try {
        } catch (e) {
            val = null;
        }`

Is it correct?

SamVerschueren:
This is another approach in order to solve #98 and PR #97. It uses [`decode-uri-component`](https://github.com/SamVerschueren/decode-uri-component) which doesn't throw an error when it fails to decode. It tries to decode as much from the input as possible. Most of the time, it uses the native `decodeURIComponent`. Only when that one fails (edge cases for instance), a custom decoder is being used.

Should we document this somewhere?

---

Fixes #98 
Fixes #97

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 var strictUriEncode = require('strict-uri-encode');
 var objectAssign = require('object-assign');
+var decodeComponent = require('decode-uri-component');
 
 function encoderForArrayFormat(opts) {
 	switch (opts.arrayFormat) {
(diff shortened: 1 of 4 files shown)


sindresorhus:
Yeah, can you add a short note at the bottom of the `.parse()` docs definition.

SamVerschueren:
I updated the docs. I came to realise though that this is actually a breaking change.

sindresorhus:
Coolio

SamVerschueren:
Nice! :)

PR #104, opened 2017-08-17
Add `.parseUrl()` method

### LINKED ISSUE #103: Support parsing full url
It would be nice to add another function, that will return an object representing the full url: url string and it's query params.
It will be using the current parse functionality.

OrRosenblatt:
Fixes #103 

Added support for parsing full url, including tests and documentation.

**Usage**
`.parseUrl(string, [options])`

Returns `Object` 
```
{
    url: '',
    queryParams: {}
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -204,3 +204,10 @@ exports.stringify = function (obj, opts) {
 		return x.length > 0;
 	}).join('&') : '';
 };
+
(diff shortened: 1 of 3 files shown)


rapzo:
I believe this has been discussed earlier.

https://github.com/sindresorhus/query-string/issues/15#issuecomment-71991610

PR #107, opened 2017-09-13
Add `sort` option

msn0:
Hi, I've added custom option to `stringify`- `compareFunction`. With this we can override default behaviour of `Array.prototype.sort`. It can also be used to disable sorting #102 in the following way

```js
stringify(urlObject, { compareFunction: () => {} });
```

What do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -172,9 +172,13 @@ exports.stringify = function (obj, opts) {
 
 	opts = objectAssign(defaults, opts);
 
+	if (opts.sort === false) {
+		opts.sort = function () {};
+	}
+
 	var formatter = encoderForArrayFormat(opts);
 
-	return obj ? Object.keys(obj).sort().map(function (key) {
+	return obj ? Object.keys(obj).sort(opts.sort).map(function (key) {
 		var val = obj[key];
 
 		if (val === undefined) {
--- a/readme.md
+++ b/readme.md
@@ -133,6 +133,31 @@ queryString.stringify({foo: [1,2,3]});
 // => foo=1&foo=2&foo=3
 ```
 
+#### sort
+
+Type: `function`
+
+Supports both `Function` as a custom sorting function or `false` for disabling sorting.
+
+- `Function`: is a custom sorting function, such as:
+
+```js
+const order = ['c', 'a', 'b'];
+queryString.stringify({ a: 1, b: 2, c: 3}, {
+	sort: (m, n) => order.indexOf(m) >= order.indexOf(n)
+});
+// => 'c=3&a=1&b=2'
+```
+
+- `false`: is the way to disable sorting, i.e.
+
+```js
+queryString.stringify({ b: 1, c: 2, a: 3}, {sort: false});
+// => 'c=3&a=1&b=2'
+```
+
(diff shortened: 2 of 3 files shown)


readme.md
```diff
@@ -133,6 +133,21 @@ queryString.stringify({foo: [1,2,3]});
 // => foo=1&foo=2&foo=3
 ```
 
+#### compareFunction
+
+Type: `function`<br>
+Default: `'none'`
```
kevva: If the default is `undefined` you can remove this line.

msn0: ok

SamVerschueren:
Same question as in #11. Why? Can you provide a use case? I can't come up with a reason why order is important.

msn0:
The reason was not even obvious for me so I decided to ask SEO team in my company.

In general, from SEO point of view, url’s with different order of query params are two different url’’s. Some query params (not all of them) are used to construct canonical url. If two content-equivalent url’s got indexed, then Google will penalize us because of duplicate content. If multiple versions of the same site will get crawled and indexed, then (besides duplicate content) we’ll run out of crawling budget. So query params order needs to be preserved for content-equivalent sites. Unfortunately they’re not always in alphabetical order because of UX point of view - some of them are more important for the user and need to be presented first, some of them are not important at all, and they’re in the end of the url. Unfortunately intersection of ux-query-params and seo-query-params is non empty.

Here’s the working example - https://allegro.pl/kategoria/twarz-podklady-45668?marka=Revlon&string=colorstay&wlasciwosci=nawilżające&produkt-wodoodporny=tak&faktor-spf=11-20&konsystencja=kremowa&dostawa-z-polski=Tak&order=m&bmatch=ss-base-relevance-floki-5-nga-hc-fas-1-2-0905

Compare query-params in url and canonical-url and note the order in the latter. Query-params are splitted into groups and only some of them are in alphabetical order.

We have our own canonical-url-builder solution which is not as efficient as we expect. Right now we try to build a new one, lighter and faster.

sindresorhus:
I'm ok with this, but the option should be named `sort` and also support `false` as a value to completely turn off sorting.

msn0:
Great, I'll fix it and let you know.

ajhsu:
The order of query string does matter to me because of the URL readability.

While some of query params were required, some were optional; Those required fields should be placed in higher order, like the following example: 

`/resource?query=value&limit=10`

That `query` is required field but `limit` is totally optional. But it will be stringified into following order: 

`/resource?limit=10&query=value`

Which is loses its readability (to separate from required and optional fields) and modifiability.

sindresorhus:
Ping @msn0 :)

index.js
```diff
@@ -172,9 +172,13 @@ exports.stringify = function (obj, opts) {
 
 	opts = objectAssign(defaults, opts);
 
+	if (opts.sort === false) {
+		opts.sort = function () {};
```
msn0: I'm not so happy about this mutation, but well... if you have a better idea, then I'm open to change it.

msn0:
@sindresorhus, @kevva, @SamVerschueren please proceed with the review.

msn0:
@sindresorhus any chance for this change?

sindresorhus:
@neuhaus In V8 the order is predictable for alphanumeric keys, so it's fine.

PR #109, opened 2017-09-20
Fix `.extract()` for query strings containing `?`

neykov:


index.js
```diff
@@ -114,7 +114,13 @@ function keysSorter(input) {
...
 
 exports.extract = function (str) {
-	return str.split('?')[1] || '';
+	var queryStart = str.indexOf('?');
+	if (queryStart == -1) {
+		// no query string
+		return '';
+	} else {
+		return str.substring(queryStart+1)
+	}
```
neykov: Even more logic would be needed if the library expects to work with URLs containing hashes (`#`).

index.js
```diff
@@ -114,7 +114,13 @@ function keysSorter(input) {
 }
 
 exports.extract = function (str) {
-	return str.split('?')[1] || '';
+	var queryStart = str.indexOf('?');
+	if (queryStart == -1) {
+		// no query string
+		return '';
+	} else {
+		return str.substring(queryStart+1)
```
sindresorhus: Can you use `.slice()` instead here. Just my preference.

index.js
```diff
@@ -114,7 +114,13 @@ function keysSorter(input) {
 }
 
 exports.extract = function (str) {
-	return str.split('?')[1] || '';
+	var queryStart = str.indexOf('?');
+	if (queryStart == -1) {
+		// no query string
```
sindresorhus: Moot comment. The code is already clear enough.

index.js
```diff
@@ -114,7 +114,13 @@ function keysSorter(input) {
 }
 
 exports.extract = function (str) {
-	return str.split('?')[1] || '';
+	var queryStart = str.indexOf('?');
+	if (queryStart == -1) {
```
SamVerschueren: Use `===`

index.js
```diff
@@ -114,7 +114,13 @@ function keysSorter(input) {
...
+		// no query string
+		return '';
+	} else {
```
SamVerschueren: Do not use the `else`

```
if (queryStart === -1) {
    return '';
}

return str.slice(queryStart+1);
```

neykov:
Addressed comments.

index.js
```diff
@@ -114,7 +114,11 @@ function keysSorter(input) {
 }
 
 exports.extract = function (str) {
-	return str.split('?')[1] || '';
+	var queryStart = str.indexOf('?');
+	if (queryStart === -1) {
+		return '';
+	}
+	return str.slice(queryStart+1);
```
kevva: Could do `return queryStart === -1 ? '' : str.slice(queryStart + 1);`

apuntovanini:
Any plans to merge this? Thanks!

rapzo:
I would if i could :-)

jorrit:
Wouldn't `return str.split('?', 2)[1] || '';` have sufficed?

PR #123, opened 2018-02-08
Add property based test to check parse reverses stringify correctly

dubzzz:


sindresorhus:
What's the goal of this? What is the benefit over the existing tests?

dubzzz:
The idea of this test is to cover (and maybe discover) edge-cases of the query-string package we have not think about in the already existing unit-tests.

**Limitation of query-string I found with this technic**

For instance, in the documentation nothing states precisely that not all the strings are valid as keys and/or values. Using this approach, I caught the following corner-case:

```javascript
const qs = require('query-string');
qs.stringify(data);
```

https://runkit.com/dubzzz/bug-query-string

Either a bug or an expected case (surrogates characters) but I have seen nothing about it in the documentation. Note that those values of string are valid for JSON.stringify for instance.

**The test itself**

Basically the test just generates a valid entry for stringify of query-string, stringify it and check that it can decode it back to its original value. It covers the range of all possible combinations and can in some cases find edge-cases difficult to diagnose and think about.

**Property based testing in a nutshell**

Property based testing is a testing approach which has appeared mostly in functional world through QuickCheck framework (Haskell). The idea behind this approach is that it is difficult to cover the range of all possible failures and corner-cases using unit-tests which are mostly example based tests in the industry.

A property is of the form: Given any entry satisfying requirements (here: an object with string keys and string values) the following statement holds (here: I can stringify it and read it back).

Thsi approach is quite useful as it makes very few assumptions on the inputs. For instance it does not state that the string should be X char longs or that the object should have X keys. Moreover whener it fails it shrink its failing input to the minimal failing one (cf. runkit link where the framework give back an example with a sign key object and a string of length one).

dubzzz:
Except the issue with half surrogate pairs, I have not found any other issues. 
I will cancel this pull request as it was not able to discover cases not handled by classical unit-tests.

PR #124, opened 2018-02-08
Add decode option to parse

yadielar:
Adds ability to pass { decode: false } to parse (the inverse of the encode option available in stringify).


yadielar:
Fixes #124

PR #131, opened 2018-03-15
Clarify what features "modern browsers" support

### LINKED ISSUE #130: Indicate what "modern browsers" means
The new release of query-string "targets Node.js 6 or later and modern browsers." It's not quite clear what "modern browsers" mean though - which versions, or what features should they support? (And probably most important: are IE and mobile Safari good enough?) It would be nice if this could be indicated in the README.

Vinnl:
I scanned through [the diff](https://github.com/sindresorhus/query-string/compare/v5.1.1...v6.0.0) of the latest release, and I think `Object.assign` and arrow functions are the primary discriminant between supported and unsupported browsers. I added that to the README so others don't have to read through the diff :)

Fixes #130.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -15,7 +15,7 @@
 $ npm install query-string
 ```
 
-The latest version targets Node.js 6 or later and modern browsers. If you want support for older browsers, use version 5: `npm install query-string@5`.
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't want to be held to such specifics. I'm gonna use any features/syntax available in the latest browsers. I've improved the readme text.

Vinnl:
Thanks, good enough :) Edge is not a consideration either?

PR #138, opened 2018-04-01
Fix stringify for bracket mode with arrays containing null

dubzzz:
I ran into an unexpected behaviour while running property based tests against query-string.

The unexpected behaviour is the following: when the settings `bracket` is enabled, null values contained inside arrays are badly exported causing problems on parsing side
```
m.stringify({bar: ['a', null, 'b']}, {arrayFormat: 'bracket'}) //=> "bar[]=a&bar&bar[]=b"
m.parse('bar[]=a&bar&bar[]=b', {arrayFormat: 'bracket'}) //=> {bar: [null, 'b']}
```

The commit contains both:
- the fix to serialize arrays correctly when enabling `bracket` mode
- the unit tests to assess the behaviour is correct
- the original property based test that detected the problem

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ function encoderForArrayFormat(options) {
 			};
 		case 'bracket':
 			return (key, value) => {
-				return value === null ? encode(key, options) : [
+				return value === null ? [encode(key, options), '[]'].join('') : [
 					encode(key, options),
 					'[]=',
 					encode(value, options)
--- a/package.json
+++ b/package.json
@@ -39,6 +39,8 @@
   },
   "devDependencies": {
     "ava": "*",
+    "deep-equal": "^1.0.1",
+    "fast-check": "^0.0.13",
     "xo": "*"
   }
 }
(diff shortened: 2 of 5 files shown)


dubzzz:
@sindresorhus 
I believe that it would be great to fix this issue. Fix in this very same PR, #138.
Without the fix parsing fails to read some query params it has itself generated.

Thanks in advance for your feedback

sindresorhus:
Any idea why the fastcheck test fails on Node.js 10? https://travis-ci.org/sindresorhus/query-string/jobs/374050952

dubzzz:
Submitting a fix for this issue as soon as possible.
In my original test, I used the wrong type for the `sort` parameter.

dubzzz:
In the log we can see that the failing case is:
- `obj = {}`
- `opts = {"arrayFormat":"bracket","strict":false,"encode":false,"sort":true}`

And that it caused the following error: `TypeError: The comparison function must be either a function or undefined` in `sort`

PR #139, opened 2018-04-09
Fixed issue with using encodeURIComponent instead if encodeURI in loose mode

alex-shamshurin:
`stringify` functions must use 'encodeURI' encoding in loose mode ( `{strict: false}`) instead of encodeURIComponent which is for strict mode. This fix adds a possibility to use dates in query string without encoding slash (/). Some web servers need additional setup to support encoded dates like `date=28%2F09%2F1979` instead of `date=28/09/1979`

### DIFF
--- a/index.js
+++ b/index.js
@@ -90,7 +90,7 @@ function parserForArrayFormat(options) {
 
 function encode(value, options) {
 	if (options.encode) {
-		return options.strict ? strictUriEncode(value) : encodeURIComponent(value);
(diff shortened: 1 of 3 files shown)


sindresorhus:
No thanks. I don't want to diverge further from the browser `URLSearchParams` API.

```
const s = new URLSearchParams({date: '28/09/1979'})
s.toString()
//=> "date=28%2F09%2F1979"
```

PR #141, opened 2018-05-01
Fix es5

samuelmanzanera:
After a failure when I build with create-react-app, I notice the package was not es5 compliant.
So i refactor to adapt for es5 modules


SamVerschueren:
Use version 5 of this module https://github.com/sindresorhus/query-string/blob/master/readme.md#install

PR #142, opened 2018-05-02
Fix failing tests in Node 10 caused by wrong sort option type

dubzzz:
The `sort` parameter used in tests added #138 was using the wrong type: boolean instead of false or function.

It caused an issue when running tests against node 10 in travis.


dubzzz:
@sindresorhus Sorry for this problem. It seems that calling sort with a `true` as a comparison function was not causing any issues before Node 10. Just fixed the tests to have the correct entries.

sindresorhus:
👌

PR #144, opened 2018-05-03
fix problem with es6 code.

wespickett:
Using es6 code as the main file was causing build issues. Switched main file to be commonjs


SamVerschueren:
Duplicate of #141

PR #146, opened 2018-05-29
Added configuration and dependencies to compile ES5 version

doktorshu:


sindresorhus:
This module mainly targets Node.js, not the browser. It's up to you to transpile it with Babel if you want to use it in the browser. You can find a more detailed explanation here: https://github.com/sindresorhus/ama/issues/446

PR #147, opened 2018-05-31
Support stringifying objects into querystrings 

k1sul1:
For whatever reason, I need to send these kinds of query strings to a server:

`_embed=1&settings[prefetch]=false&settings[expiry]=3600`

I figured I'd use a similar data structure; a JavaScript object, and was a bit annoyed when stringifying `{settings: { prefetch: false, expiry: 3600 }}` resulted in an url encoded [object object].

This fixes that. I added a new arrayFormat option, `key`. 

I tried giving a shot at `parse`, but we need one way conversion only at this point, and I couldn't figure it out in 10 minutes, so I gave up. Regular expressions are incomprehensible to me.

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,6 +27,14 @@ function encoderForArrayFormat(options) {
 					encode(value, options)
 				].join('');
 			};
+		case 'key':
+			return (parent, key, value) => {
+				return value === null ? [encode(parent, options), '[' + key + ']'].join('') : [
+					encode(parent, options),
+					'[' + key + ']=',
+					encode(value, options)
+				].join('');
+			};
 		default:
 			return (key, value) => {
 				return value === null ? encode(key, options) : [
@@ -208,6 +216,14 @@ exports.stringify = (obj, options) => {
 			}
 
 			return result.join('&');
+		}
+
+		if (typeof value === 'object') {
+			return Object.entries(value).reduce((acc, [k, v]) => {
+				acc.push(formatter(key, k, v));
+
+				return acc;
+			}, []).join('&');
 		}
 
 		return encode(key, options) + '=' + encode(value, options);
--- a/test/stringify.js
+++ b/test/stringify.js
@@ -117,6 +117,23 @@ test('array stringify representation with array brackets and null value', t => {
 	}), 'bar[]&foo[]=a&foo[]&foo[]=');
 });
 
+test('array stringify representation with array keys', t => {
+	t.is(m.stringify({
+		foo: null,
+		bar: {one: 'two', three: 'four'}
+	}, {
+		arrayFormat: 'key'
+	}), 'bar[one]=two&bar[three]=four&foo');
+});
+
+test('array stringify representation with array keys and null value', t => {
+	t.is(m.stringify({
+		foo: {a: null, b: 'c'}
+	}, {
+		arrayFormat: 'key'
+	}), 'foo[a]&foo[b]=c');
+});
+
 test('array stringify representation with a bad array format', t => {
 	t.is(m.stringify({
 		foo: null,


k1sul1:
Same test appears to fail in master.

Chieftl:
Can you tell me please, how about nested objects?
`{settings: { prefetch: false, expiry: {endDate: 5} }}`

Result looks like:
`settings[prefetch]=false&settings[expiry]=[object Object]`

k1sul1:
Same behaviour .

https://github.com/sindresorhus/query-string#nesting

k1sul1:
I hate query strings.

Mixing objects and arrays doesn't quite work. 

```
{
  "settings": {
    "prefetch": true,
    "expiry": 3600
  },
  "categories": [
    1057
  ],
}
```

results in 

`categories[1057]=0&settings[prefetch]=true&settings[expiry]=3600`

when it should be 

`categories[]=1057&settings[prefetch]=true&settings[expiry]=3600`

I'm just going to have to parse things manually and append them to the querystring generated by this library, there are way too many edge cases to handle and no time at all to do so.

Chieftl:
### Here were go, my solution:

1. Install [q-flat](https://www.npmjs.com/package/q-flat) from npm
This module convert nested objects to flat

Example:
```
{
  "page": {
    "limit": 10,
    "offset": 55
  },
  "filter": {
    "year": {
      "gt": 2000
    }
  },
  "scalar": 3.1415,
  "categories": [
    1057
  ]
}
```
will be convert to
```
{
  "page[limit]": 10,
  "page[offset]": 55,
  "filter[year][gt]": 2000,
  "scalar": 3.1415,
  "categories[0]": 1057
}
```

2. Import q-flat & use it on parameters
```
import queryString from 'query-string'
import { flatten } from 'q-flat'

var paramsFlat = flatten(params);
var urlQuery = queryString.stringify(paramsFlat, {encode:false});
// categories[0]=1057&filter[year][gt]=2000&page[limit]=10&page[offset]=55&scalar=3.1415
```

Voilà!

bogdansoare:
@Chieftl how would you parse that query flatten with q-flat?

sindresorhus:
I'm gonna pass on this. Supporting objects is just too messy. Check out https://github.com/ljharb/qs instead.

bogdansoare:
@sindresorhus then what are the advantages of this library over [ljharb/qs](https://github.com/ljharb/qs)?

sindresorhus:
Not supporting objects is a feature. `qs` has had multiple vulnerabilities in the past and its codebase is more complicated and it has more bugs.

bogdansoare:
Understood, thank you for the clarification 🙏

Chieftl:
@bogdansoare @sindresorhus Yes, [qs](https://github.com/ljharb/qs) is good choise for unflatting if you are using NodeJS on backend.
My code:
```javascript
var http = require('http');
var queryString = require('query-string');
var qs = require('qs');

http.createServer(function (req, res) {
    var query = req.url.replace(/^\/\?/, '');
    var queryObj = qs.parse(query);

    var report = [
        query,
        JSON.stringify(queryObj),
    ].join("\n");

    res.end(report);
}).listen(3333);

console.log("Server running at http://localhost:3333/");
```

Url:

http://localhost:3333/?categories[0]=1057&filter[year][gt]=2000&page[limit]=10&page[offset]=55&scalar=3.1415

Result:
`
categories[0]=1057&filter[year][gt]=2000&page[limit]=10&page[offset]=55&scalar=3.1415
{"categories":["1057"],"filter":{"year":{"gt":"2000"}},"page":{"limit":"10","offset":"55"},"scalar":"3.1415"}
`

PR #148, opened 2018-06-22
Doesn't support webpack

Dorian:
Failed to minify the code from this file:

 	./node_modules/query-string/index.js:8

Read more here: http://bit.ly/2tRViJ9

--------

I couldn't post an issue, so I make a pull request instead

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,6 +2,16 @@
 
 > Parse and stringify URL [query strings](https://en.wikipedia.org/wiki/Query_string)
 
+Doesn't support webpack:
+
+```
+Failed to minify the code from this file:
+
+ 	./node_modules/query-string/index.js:8
+
+Read more here: http://bit.ly/2tRViJ9
+```
+
 ---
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's the inverse, actually. It's Create React App that doesn't support the ecosystem. Use `npm install query-string@5` if you use CRA. Sorry for the trouble, but this is all on CRA which has refused to co-operate with the community and just tells users to annoy maintainers.

> I couldn't post an issue, so I make a pull request instead

The issue tracker is disabled exactly because of issues like this.

Dorian:
No problem, thanks for the answer.

For anyone reading this in the future, I replaced the `query-string` package by this function in `src/query-string.js`:

```
export default {
  parse: function parse(queryString) {
    var query = {};
    var pairs = (queryString[0] === '?' ? queryString.substr(1) : queryString).split('&');
    for (var i = 0; i < pairs.length; i++) {
        var pair = pairs[i].split('=');
        query[decodeURIComponent(pair[0])] = decodeURIComponent(pair[1] || '');
    }
    return query;
  }
}
```

jayantbh:
This https://github.com/sindresorhus/query-string/pull/148#issuecomment-399656020 should actually be added to the README.md. I imagine it would be a fairly common problem.

slorber:
@sindresorhus what should the CRA community do about this error exactly? I've had it also on a non-CRA project (actually, Ruby/Webpacker env) on which I control webpack settings, so wonder what's your advise on how to keep up to date with this package?

Adding config to transpile this module seems a bit annoying so I'm sticking to 5.x for now, just wondering what are the alternatives

slipo:
Agree with @jayantbh. PR https://github.com/sindresorhus/query-string/pull/156 created.

PR #150, opened 2018-08-08
Add single parameter stringify option for array format

rcoy-v:
This adds a new option for `arrayFormat` that allows an array to be stringified as a single query parameter. An example of using this format is if an application/WAF does not allow multiple declarations of a query parameter to limit possibilities of a payload becoming malicious once concatenated by the receiving API.

These are a few links further describing HTTP parameter pollution:
https://www.owasp.org/index.php/Testing_for_HTTP_Parameter_pollution_(OTG-INPVAL-004)
https://tacticalwebappsec.blogspot.com/2009/05/http-parameter-pollution.html
https://capec.mitre.org/data/definitions/460.html
https://medium.com/@aneeskhan/how-i-bypassed-open-redirection-tokens-using-http-parameter-pollution-ea5be084a0f4

### DIFF
--- a/index.js
+++ b/index.js
@@ -197,6 +197,16 @@ exports.stringify = (obj, options) => {
 		}
 
 		if (Array.isArray(value)) {
+			if (options.arrayFormat === 'single') {
+				return [
+					encode(key, options),
+					'=[',
+					(value.length > 0 ? value.reduce((acc, current, index) =>
+						index ? [acc, ',', encode(current, options)].join('') : [acc, encode(current, options)].join(''), '') : ''),
+					']'
+				].join('');
+			}
+
 			const result = [];
 
 			for (const value2 of value.slice()) {
--- a/index.js
+++ b/index.js
@@ -76,6 +76,16 @@ function parserForArrayFormat(options) {
 
 				accumulator[key] = [].concat(accumulator[key], value);
 			};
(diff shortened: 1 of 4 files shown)


test/properties.js
```diff
@@ -29,5 +29,5 @@ const optionsArbitrary = fc.record({
 test('should read correctly from stringified query params', t => {
 	t.notThrows(() => fc.assert(
 		fc.property(queryParamsArbitrary, optionsArbitrary,
-		(obj, opts) => deepEqual(m.parse(m.stringify(obj, opts), opts), obj))));
+			(obj, opts) => deepEqual(m.parse(m.stringify(obj, opts), opts), obj))));
```
rcoy-v: I will rebase this pull request once #149 is merged.

rcoy-v:
Because of the lack of standards around defining a list within a single query param, I don't feel comfortable adding this to such a re-usable library. There is [RFC 6570](https://tools.ietf.org/html/rfc6570#section-3.2.8) with guidelines, but there are also discrepancies among common language frameworks, such as https://github.com/hapijs/joi/issues/570. I'm not confident enough to say that this pull request is "standard", rather than a custom implementation for a specific use-case.

I still pushed up an implementation for `parse` as well, in case someone finds a use for this in the future.

PR #151, opened 2018-08-15
Support hash in `parseUrl()`

tranvansang:


index.js
```diff
@@ -215,8 +215,15 @@ exports.stringify = (obj, options) => {
 };
 
 exports.parseUrl = (input, options) => {
+	let hash = '';
+	const qSplits = input.split('?').map(part => {
+		const sSplits = part.split('#');
```
sindresorhus: Use array destructuring to name the parts to improve code readability.

index.js
```diff
@@ -215,8 +215,15 @@ exports.stringify = (obj, options) => {
...
+		hash = sSplits[1] || hash;
+		return sSplits[0];
+	});
 	return {
-		url: input.split('?')[0] || '',
-		query: parse(extract(input), options)
+		url: qSplits[0],
+		hash,
```
sindresorhus: The hash should still just be part of the URL and not an exposed property.

tranvansang: IMHO, Exposing hash makes comparability with `{pathname, hash, search}` pattern in other lib like [react-router](https://medium.com/@pshrmn/a-simple-react-router-v4-tutorial-7f23ff27adf)

sindresorhus: The point of the method is to allow you to get the query from an URL. I don't want to overload it with lots of other URL parsing stuff. It would also be a breaking change, as people might depend on getting the hash in the `url` property, and I'm not interested in doing a major release just for this.

elias6:
Thank you. Does this fix work in IE? If not, can it (realistically) be backported to version 5?

waynebloss:
This was a breaking change. In version `6.1.0`, the following was happening (js formatted for [RunKit](https://npm.runkit.com/query-string).)

```js
var qs = require("query-string@6.1.0");
qs.parseUrl("https://www.site.com/#/spa-resource?x=1234");
// Result:
// Object {query: Object {x: "1234"}, url: "https://www.site.com/#/spa-resource"}
```

With `6.2.0`:

```js
var qs = require("query-string@6.2.0");
qs.parseUrl("https://www.site.com/#/spa-resource?x=1234");
// Result:
// Object {query: Object {}, url: "https://www.site.com/"}
```

PR #152, opened 2018-08-16
Update readme.md

marxjb:
the correct option is "brackets" not "bracket"


marxjb:
you could keep "bracket" in the explanation, but the option.arrayFormat must be "brackets" ;)

sindresorhus:
Uhm, no: https://github.com/sindresorhus/query-string/blob/b4dee485d7f98034aa757d7c5d17b04ce73d2855/index.js#L22

PR #153, opened 2018-08-29
Changed for ... of loops to use 'let' as 'const' breaks in firefox

brunocangs:
I have been using the package with webpack for react, and it breaks on firefox because it says 'missing = in const declaration'. This seems to fix it

### DIFF
--- a/index.js
+++ b/index.js
@@ -144,7 +144,7 @@ function parse(input, options) {
 		return ret;
 	}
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's up to you to transpile it with Babel if you want to use it in older browsers. You can find a more detailed explanation here: https://github.com/sindresorhus/ama/issues/446

PR #154, opened 2018-09-06
Bump fast-check version and adapt properties

dubzzz:
Content of this review:
- bump of fast-check version
- move from `fc.object` - able to construct deep objects - to `fc.dictionary` - more adapted for this case
- use of `fc.array(arbitrary, minSize, maxSize)` for better array generation performances
- fix the property code that allowed `encode: false` case - *PR https://github.com/dubzzz/fast-check/pull/175 (not merged  yet) enhances the corner case detection on strings and report an error for this case*

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
 	"devDependencies": {
 		"ava": "*",
 		"deep-equal": "^1.0.1",
-		"fast-check": "^1.0.1",
+		"fast-check": "^1.5.0",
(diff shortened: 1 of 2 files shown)


sindresorhus:
Should I wait on merging this until https://github.com/dubzzz/fast-check/pull/175 is merged and released?

dubzzz:
No you can merge it right now

PR #155, opened 2018-09-09
Do not sort when `sort` option is set to false

vicapow:
We cannot rely on `() => {}` as a no-op for sorting.


vicapow:
For the most part, this should be backwards compatible unless you were setting `sort: false` and happen to be encoding parameters that were susceptible to this issue.

vicapow:
anything else you'd like me to tackle for this?

PR #156, opened 2018-09-15
Add readme note about `create-react-app`

slipo:
It's confusing if you're using create-react-app and trying to add query-string in. [As jayantbh mentioned](https://github.com/sindresorhus/query-string/pull/148#issuecomment-411693303) we should note this in the README with the known solution.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -15,7 +15,7 @@
 $ npm install query-string
 ```
 
-This module targets Node.js 6 or later and the latest version of Chrome, Firefox, and Safari. If you want support for older browsers, use version 5: `npm install query-string@5`.
+This module targets Node.js 6 or later and the latest version of Chrome, Firefox, and Safari. If you want support for older browsers, or, [if your project is using create-react-app](https://github.com/sindresorhus/query-string/pull/148#issuecomment-399656020), use version 5: `npm install query-string@5`.
 
 <a href="https://www.patreon.com/sindresorhus">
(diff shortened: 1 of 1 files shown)


psyanite:
I was running 6.1.0, have downgraded to 5 something and is building perfectly now.

jamesplease:
More broadly, @sindresorhus , would you be open to reverting commit  597f14a1c39b16d4b879c3834b006884782c851b ?

I don't think that the value that it provides offsets the troubles that it adds when trying to consume the library. Consumers don't care whether this uses `const` or `var` under the hood – they just want something that works. And, right now, the latest version doesn't work for a lot of people, and that is what they get when they run `npm i query-string`.

For me, any time that I start a new project, I find myself installing v6 (because that's what from `npm i query-string`), getting confused for a moment when my build system explodes, then remembering that I need _v5_, and finally switching over.

Alrigh, so maybe that's not a big deal. I install v5, and then I'm good, right? Not quite, as I get warnings that my dependencies are out of date for the _entire lifetime_ of the project.

sindresorhus:
> More broadly, @sindresorhus , would you be open to reverting commit 597f14a ?

No, sorry.

sindresorhus:
[Create React App v2](https://reactjs.org/blog/2018/10/01/create-react-app-v2.html) supports transpilation of dependencies, so this will not be a problem for much longer.

PR #157, opened 2018-10-09
set default arrayFormat

zhangciwu:
Add the ability to set an arrayFormat as global default setting
convenient for all same setting in one project.

like `jQuery.ajaxSetup()`

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 const strictUriEncode = require('strict-uri-encode');
 const decodeComponent = require('decode-uri-component');
+
 let DEFAULT_ARRAY_FORMAT = 'none';
 
 function encoderForArrayFormat(options) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
This is a very bad idea. It would also affect any usage elsewhere in the dependency tree. Would be better to add support for creating an instance: `queryString.create({encodeOptions: …, decodeOptions: …})`, but not sure it's worth the bloat.

zhangciwu:
@sindresorhus It is an issue that affecting other libs.
But maybe it's what I want, especially that libs not caring about this setting. (just give no option when they use query-string)

Maybe add a description to make lib writer pass option if they intend to not be affected by user default. (which shall be done if they have use other than `none`)

sindresorhus:
I'm not going to add a bad solution even with good docs.

zhangciwu:
Agree, let me think if there is a better solution

PR #158, opened 2018-10-23
Support various targets

bwendt-mylo:
I wasn't paying attention and installed the latest 6.x version instead of 5.x, which didn't properly transpile through webpack. I'm not the only one: https://stackoverflow.com/questions/49954618/babel-not-transpiling-arrow-functions-webpack

I think this would be a useful addition.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+dist/
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "query-string",
-	"version": "6.2.0",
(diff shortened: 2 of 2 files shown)


sindresorhus:
Thanks for the suggestion, but this is not something I want.

jolyonruss:
As someone who's PROD build broke for legacy browsers due to upgrading query-string to version 6.x - I'm coming here to second @bwendt-mylo's assertion this would be a useful addition.

This is something that I want.

benjaminkniaz:
Same issue here.

bj00rn:
https://github.com/sindresorhus/query-string/pull/159

PR #159, opened 2018-12-20
Use Babel to compile to ES5.

cdeutsch:
Any interest in this or should I publish a `query-string-es5` NPM?

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["@babel/preset-env"]
+}
--- a/dist/index.js
+++ b/dist/index.js
@@ -1,0 +1,277 @@
+'use strict';
+
+function _slicedToArray(arr, i) { return _arrayWithHoles(arr) || _iterableToArrayLimit(arr, i) || _nonIterableRest(); }
+
+function _nonIterableRest() { throw new TypeError("Invalid attempt to destructure non-iterable instance"); }
+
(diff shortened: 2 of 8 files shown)


sindresorhus:
No interest

cdeutsch:
Why even spend the time to contribute your code if you're not willing to do the simplest things to support web developers 👎

SamVerschueren:
This module mainly targets Node.js, not the browser. It's up to you to transpile it with Babel if you want to use it in the browser. You can find a more detailed explanation here: https://github.com/sindresorhus/ama/issues/446

cdeutsch:
> This module mainly targets Node.js

How so? There is nothing specific to Node about this code. Don't you think browsers need to deal with query strings? Not to mention it's barely even using ES6. Just arrow functions and `const` and it's a single file. It feels like an intentional middle finger to web developers.

I've read @sindresorhus  *opinion* on the matter of jumping to ES6. Unfortunately for web developers Microsoft bundled IE11 with Windows 10, so we're stuck supporting an ES5 only browser until 2025.
https://medium.com/@burger.neal/the-end-of-life-of-internet-explorer-11-12736f9ff75f

I get he doesn't have any empathy for web developers because he doesn't do browser development, but this decision has come back to bite me and many other developers in the ass, and wasted countless hours of people's time.

The NPM ecosystem and frontend build systems weren't ready for this change, when he started shipping ES6 packages. Would have been nice to have giant warnings about not using it on the web at the top of the README, instead of marketing affiliate links. 👎 

Thanks for the link to `babel-engine-plugin`, I'll look into that. 👍  

I spent a whole day looking for solutions that are better than running the whole of `node_modules` through Babel and couldn't find any, and my team has been complaining ever since about slower builds.

iamvijaydev:
This PR should have been accepted and merged. I feel you @cdeutsch

amqqliang:
This package provides great convenience for our developers in some scenarios. Moreover, there are often similar requirements in browser, just like my project. Sincerely hope that this package can support ES5, then the universality of this package is even higher.

cdeutsch:
You can use [query-string-for-all](https://github.com/cdeutsch/query-string-for-all) which I keep in sync with this library.

Isn't the JavaScript ecosystem great? 🙄

PR #160, opened 2018-12-26
function parseUrl add new option isSupportHashQuery

itwara:
parseUrl add a new option  'isSupportHashQuery'


sindresorhus:
Can you elaborate on what this PR adds and why it's needed?

PR #162, opened 2019-01-10
Add `sort` option to the `.parse()` method

sachaudh:
*Reason for change:* 

I have a use case for a Search Input component I'm making that needs to interact with the URL as a source of truth for any filtering that needs to be done in my UI. The user would typically go through the following scenario:

1. Search input currently has the following chips: `Cluster:` `DockerSwarm` `Namespace:` `test` and the corresponding search query is `?Cluster=DockerSwarm&Namespace=test`
2. If they add another search term for "Deployment=nginx" then the chips show up as `Cluster:` `DockerSwarm` `Deployment:` `nginx` `Namespace:` `test`.  The corresponding search query would be `?Cluster=DockerSwarm&Namespace=test&Deployment=nginx`. The `query-string` library will sort the keys alphabetically when it parses the query string, and that causes inconsistencies between the order you see in the search query and the order of the chips in the search input.

To prevent things from jumping around due to the sorting, I thought i'd add an option to prevent sorting when we pass `{sort: false}`. By default it'll be true.

Let me know if this makes sense. Thought i'd send a PR 🤷‍♂️

### DIFF
--- a/index.js
+++ b/index.js
@@ -128,7 +128,7 @@ function extract(input) {
 }
 
 function parse(input, options) {
-	options = Object.assign({decode: true, arrayFormat: 'none'}, options);
+	options = Object.assign({decode: true, arrayFormat: 'none', sort: true}, options);
 
 	const formatter = parserForArrayFormat(options);
 
@@ -153,6 +153,10 @@ function parse(input, options) {
 		value = value === undefined ? null : decode(value, options);
 
 		formatter(decode(key, options), value, ret);
+	}
+
+	if (!options.sort) {
+		return ret;
 	}
 
 	return Object.keys(ret).sort().reduce((result, key) => {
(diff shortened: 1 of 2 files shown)


idMolotov:
@sindresorhus please review

mogelbrod:
I drafted a updated documentation: https://github.com/sindresorhus/query-string/commit/50eddef5b3848648e21c83a36c7cb4059019ca5f

The `sort` option for `parse()` should probably mirror the one for `stringify()` though, and accept both booleans and functions.

sindresorhus:
> The sort option for parse() should probably mirror the one for stringify() though, and accept both booleans and functions.

👍

sindresorhus:
You also need to update https://github.com/sindresorhus/query-string/blob/master/index.d.ts

sindresorhus:
Bump

idMolotov:
@sachaudh please, finish the changes. You just need add definition into the .ts file sachaudh the same as here `https://github.com/sindresorhus/query-string/blob/master/index.d.ts#L129`.

PR #164, opened 2019-01-22
Use babel for IE browsers

steelydylan:
Thank you for the awesome works!
It's really convenient!

I know that this module is for Node.js but I also want to use it for IE browsers without changing webpack.config.js.
Please check the PR ><


sindresorhus:
Duplicate of #159

PR #167, opened 2019-03-01
Add `comma` format to the `arrayFormat` option

zepod:
This pull requests adds fourth `arrayFormat` option, named `comma`. We needed this for our project, as we needed to support arrays in query in below shape...

`foo=1,2,3,4,5,6,7,8`

This format is mainly used to shorten URLs which include things as identifiers, where there is expected to be a lot of them.

 On the way to add this feature I had to slightly reimplement `encoderForArrayFormat` to use `Array.prototype.reduce` instead of previous `for .. in`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -29,7 +29,7 @@ export interface ParseOptions {
 	 *    queryString.parse('foo=1&foo=2&foo=3');
 	 *    //=> foo: [1,2,3]
 	 */
-	readonly arrayFormat?: 'bracket' | 'index' | 'none';
+	readonly arrayFormat?: 'bracket' | 'index' | 'comma' | 'none';
 }
 
 export interface ParsedQuery {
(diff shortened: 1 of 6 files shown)


tomaass:
Great work @zepod !
Fix tests please I really want to use it on my project 🥇

sindresorhus:
Why did you have to switch to Array#reduce?

sindresorhus:
You have to include TypeScript types for the changes too, since a TypeScript declaration file was recently added to `master`.

zepod:
@sindresorhus 
the way it was implemented with `for` `encoderForArrayFormat` was a unary function mapping **single element** of provided array, to **single key** in the query sequence.

Using comma format, we wanted to map the **multiple elements** of array to a **single key** in the query.

zepod:
Typescript types added

sindresorhus:
Looks good. Thank you :)

PR #168, opened 2019-03-03
Add TypeScript definition

BendingBender:


index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+	encode?: boolean;
+
+	/**
+	 * Supports both `index` for an indexed array representation or `bracket` for a *bracketed* array representation.
+	 *
+	 * @default 'none'
+	 *
+	 * - `bracket`: stands for parsing correctly arrays with bracket representation on the query string, such as:
+	 *
+	 *
```
sindresorhus: Why the extra empty line here?

index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Parse a query string into an object. Leading `?` or `#` are ignored, so you can pass `location.search` or `location.hash` directly.
+ *
+ * The returned object is created with [`Object.create(null)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) and thus does not have a `prototype`.
+ *
+ * @param query - The query string to parse.
+ */
+export function parse(query: string, options?: ParseOptions): ParsedQuery;
+
+export interface ParsedUrl {
```
sindresorhus: I prefer defining interface/type before it's used.

index.d.ts
BendingBender: VS Code only renders the following example as code if I add this extra empty line.

An alternative (which I dislike) would be to write it like this:
````
	 * - `bracket`: stands for parsing correctly arrays with bracket representation on the query string, such as:
	 * 
	 * ```js
  queryString.parse('foo[]=1&foo[]=2&foo[]=3', {arrayFormat: 'bracket'});
  //=> foo: [1,2,3]
```
	 *
````

sindresorhus: Ah ok. I had a feeling it was something like that.

Cellule:
A little late now, but would have been nice for the definition file to match closer to the version in DefinitelyTyped
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/query-string/index.d.ts
Specifically `InputParams` vs `{[key: string]: unknown}` and `OutputParams` vs `ParsedQuery`
`ParsedUrl` is a welcome addition though

BendingBender:
This is a necessary evil. This one time we have to break compat with DT.

The benefit is that now types will always be up to date and if a breaking change is introduced to the types, a major version will be released.

In DT, breaking changes to types may be introduced even to atch releases.

PR #169, opened 2019-03-06
Correctly handle query strings that contain `=`

genomics-geek:
Some of our GraphQL IDs had `=` in their value, therefore splitting on `=` creates issues - unless we split by the 1st occurrence.

For example: `?testCode_In=VGVzdENvZGVOb2RlOjQ=` will be parsed as: `{testCode_In: "VGVzdENvZGVOb2RlOjQ"}`.  This leads to confusion and inaccurate search terms.

This has resolved the issue for us.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 const strictUriEncode = require('strict-uri-encode');
 const decodeComponent = require('decode-uri-component');
+const splitOnFirst = require('split-on-first');
 
 function encoderForArrayFormat(options) {
 	switch (options.arrayFormat) {
@@ -150,7 +151,7 @@ function parse(input, options) {
 	}
 
 	for (const param of input.split('&')) {
-		let [key, value] = param.replace(/\+/g, ' ').split('=');
+		let [key, value] = splitOnFirst(param.replace(/\+/g, ' '), '=');
 
 		// Missing `=` should be `null`:
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
 		"ava": "^1.2.1",
 		"deep-equal": "^1.0.1",
 		"fast-check": "^1.5.0",
+		"split-on-first": "^1.0.0",
 		"tsd-check": "^0.3.0",
 		"xo": "^0.24.0"
 	}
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -150,7 +150,7 @@ function parse(input, options) {
 	}
 
 	for (const param of input.split('&')) {
-		let [key, value] = param.replace(/\+/g, ' ').split('=');
+		let [key, value] = param.replace(/\+/g, ' ').split(/\=(.+)/); // eslint-disable-line no-useless-escape
```
sindresorhus: ESLint is correct. You don't have to escape `=`.

sindresorhus:
Can you add a test?

index.js
genomics-geek: @sindresorhus - yup I know, but then I get a different ESLint error about escaping.   Yup I will add a test.  Also, it seems to fail other tests now, so I have to look into it

genomics-geek:
I added tests and replaced the regex with a `replace().split()` - not as nice as regex, but it avoids adding extra element to the split.  Let me know what you think @sindresorhus

index.js
```diff
@@ -150,7 +150,7 @@ function parse(input, options) {
 	}
 
 	for (const param of input.split('&')) {
-		let [key, value] = param.replace(/\+/g, ' ').split('=');
+		let [key, value] = param.replace(/\+/g, ' ').replace('=', '<**>').split('<**>');
```
cam-perry: Just ran into the same issue. This is the alternative I used for a fix https://github.com/cam-perry/query-string/commit/b65a53b55ae7a8825d62af757a587d365ba7e3b7:
```js
const i = param.indexOf("=");
let key = param.slice(0, i);
let value = param.slice(i+1);
```

MatheusFelipe:
We're also experiencing this issue. When will the fix be merged?

index.js
sindresorhus: Ah ok. Better to ignore the `no-div-regex` rule.

index.js
sindresorhus: The previous regex was better. This is risky as it will fail if the query string actually contains `<**>`, even if unlikely.

sindresorhus:
You can use my https://github.com/sindresorhus/split-on-first module.

genomics-geek:
@sindresorhus - Nice.  I used your package like you suggested.  This should take care of the GraphQL ids now 👍

genomics-geek:
@sindresorhus - is there anything else that is needed in order to merge?

PR #170, opened 2019-03-09
Fix failing test and `ParsedQuery` TypeScript type

sn42:
- Fixes the failing test `should sort keys in given order` of the stringify function. The test's sort function didn't result in the specified order (expected: `c=baz&a=foo&b=bar`, actual: `a=foo&b=bar&c=baz`).
- Fixes the `ParsedQuery` type. Values may be `null` as seen [here](https://github.com/sindresorhus/query-string/blob/c138b1123462882bb2b44d8ecac83f0b372b53eb/index.js#L157).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -33,7 +33,7 @@ export interface ParseOptions {
 }
 
 export interface ParsedQuery {
-	readonly [key: string]: string | string[] | undefined;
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks :)

PR #171, opened 2019-03-15
Add `parseNumbers ` option to `.parse()`

yaodingyd:
Current `parse` would return all value as string but I would like an option to return number as number.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,6 +59,11 @@ export interface ParseOptions {
 	 * Parse the value as a number type instead of string type if it's a number.
 	 *
 	 * @default false
+	 *
+	 * @example
+	 *
+	 * queryString.parse('foo[]=1&foo[]=2&foo[]=3', {parseNumbers: true});
+	 * //=> foo: [1, 2, 3]
 	 */
 	readonly parseNumbers?: boolean;
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,25 +13,25 @@ export interface ParseOptions {
 	 *
 	 *
 	 *    queryString.parse('foo[]=1&foo[]=2&foo[]=3', {arrayFormat: 'bracket'});
-	 *    //=> foo: [1, 2, 3]
+	 *    //=> foo: ['1', '2', '3']
 	 *
 	 * - `index`: Parse arrays with index representation:
 	 *
 	 *
 	 *    queryString.parse('foo[0]=1&foo[1]=2&foo[3]=3', {arrayFormat: 'index'});
-	 *    //=> foo: [1, 2, 3]
+	 *    //=> foo: ['1', '2', '3']
 	 *
 	 * - `comma`: Parse arrays with elements separated by comma:
 	 *
 	 *
 	 *    queryString.parse('foo=1,2,3', {arrayFormat: 'comma'});
-	 *    //=> foo: [1, 2, 3]
+	 *    //=> foo: ['1', '2', '3']
 	 *
 	 * - `none`: Parse arrays with elements using duplicate keys:
 	 *
 	 *
 	 *    queryString.parse('foo=1&foo=2&foo=3');
-	 *    //=> foo: [1, 2, 3]
+	 *    //=> foo: ['1', '2', '3']
 	 */
 	readonly arrayFormat?: 'bracket' | 'index' | 'comma' | 'none';
 
@@ -55,10 +55,16 @@ export interface ParseOptions {
 	 */
 	readonly sort?: ((itemLeft: string, itemRight: string) => number) | false;
 
+	/**
+	 * Parse the value as a number type instead of string type if it's a number.
+	 *
+	 * @default false
+	 */
+	readonly parseNumbers?: boolean;
 }
 
 export interface ParsedQuery {
-	readonly [key: string]: string | string[] | null | undefined;
+	readonly [key: string]: string | number | Array<string | number> | null | undefined;
 }
 
 /**
--- a/index.js
+++ b/index.js
@@ -176,6 +176,8 @@ function parse(input, options) {
 		decode: true,
 		sort: true,
 		arrayFormat: 'none'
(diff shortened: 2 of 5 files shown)


zepod:
I may be worth it to update typescript types

index.js
```diff
@@ -188,6 +190,10 @@ function parse(input, options) {
 		// http://w3.org/TR/2012/WD-url-20120524/#collect-url-parameters
 		value = value === undefined ? null : decode(value, options);
 
+		if (options.number && isNumber(value)) {
```
sindresorhus: You don't need a module for this. Just try `Number(value)` and see if it's NaN or not.

readme.md
```diff
@@ -76,30 +76,37 @@ Default: `none`
 
 ```js
 queryString.parse('foo[]=1&foo[]=2&foo[]=3', {arrayFormat: 'bracket'});
-//=> foo: [1, 2, 3]
+//=> foo: ['1', '2', '3']
```
sindresorhus: Good catch. You need to update this in the TS definition too.

test/parse.js
```diff
@@ -198,3 +198,12 @@ test('decode keys and values', t => {
 test('disable decoding of keys and values', t => {
 	t.deepEqual(m.parse('tags=postal%20office,burger%2C%20fries%20and%20coke', {decode: false}), {tags: 'postal%20office,burger%2C%20fries%20and%20coke'});
 });
+
+test('number value returns as string by default', t => {
+	t.deepEqual(m.parse('foo=1'), {foo: '1'});
+});
+
+test('number value returns as number', t => {
+	t.deepEqual(m.parse('foo=1', {number: true}), {foo: 1});
```
sindresorhus: Add more test-cases. Floats, etc.

sindresorhus:
I don't think `number` is clear enough as an option name. It doesn't describe what it does. Need a more explicit name.

yaodingyd:
How about `parseAsNumber`?

Andrew-Lahikainen:
Would it make sense to make this option a little more generic so it can handle other primitives? Or do you think there is a possibility that someone might want number conversions and *not* boolean conversions for example.

If you want to have a generic name I think something like parseTypes or parsePrimitiveTypes.

If you want to handle each primitive separately, I think appending parse at the front might be acceptable?

E.g.:
* parseNumbers
* parseBooleans
* parseNulls
* parseUndefineds
* parseSymbols (this one probably doesn't make sense to implement)

sindresorhus:
> Would it make sense to make this option a little more generic so it can handle other primitives?

I don't see us adding more than `number` and `boolean` parsing. It can quickly get ambiguous. I would have preferred to have them under one option, but I have feeling people have use-cases for only either.

sindresorhus:
I prefer the `parseNumber` option name.

index.d.ts
```diff
@@ -13,31 +13,37 @@ export interface ParseOptions {
...
 	 * - `none`: Parse arrays with elements using duplicate keys:
 	 *
 	 *
 	 *    queryString.parse('foo=1&foo=2&foo=3');
-	 *    //=> foo: [1, 2, 3]
+	 *    //=> foo: ['1', '2', '3']
 	 */
 	readonly arrayFormat?: 'bracket' | 'index' | 'comma' | 'none';
+	/**
+	 * If value is a number, parse it to a number instead of string.
```
sindresorhus: This sentence is misleading. "if the value is a number" makes it sound like the value should be of type `number`.

```suggestion
	 * Parse the value as a number type instead of string type if it's a number.
```

Andrew-Lahikainen:
I prefer plural since this option parses all numbers, not just one. But that's being really nitpicky :p

sindresorhus:
Alright, `parseNumbers`.

HHK1:
Thanks for this option, it's awesome !
Are things missing before it can be merged ? Just to have an idea of when it will included in a release :)

index.d.ts
```diff
@@ -13,31 +13,37 @@ export interface ParseOptions {
...
 	 *
 	 *    queryString.parse('foo=1&foo=2&foo=3');
-	 *    //=> foo: [1, 2, 3]
+	 *    //=> foo: ['1', '2', '3']
 	 */
 	readonly arrayFormat?: 'bracket' | 'index' | 'comma' | 'none';
+	/**
+	 * Parse the value as a number type instead of string type if it's a number.
+	 *
+	 * @default false
```
sindresorhus: Can you include a short usage example here?

readme.md
```diff
@@ -76,30 +76,37 @@ Default: `none`
...
 ```
 
+##### parseNumbers
+
+Type: `boolean`<br>
+Default: `false`
+
+Parse the value as a number type instead of string type if it's a number.
```
sindresorhus: Can you include a short usage example here?

test/parse.js
```diff
@@ -202,3 +202,18 @@ test('decode keys and values', t => {
...
+test('number value returns as number if option is set', t => {
+	t.deepEqual(queryString.parse('foo=1', {parseNumbers: true}), {foo: 1});
+	t.deepEqual(queryString.parse('foo=12.3&bar=123e-1', {parseNumbers: true}), {foo: 12.3, bar: 12.3});
+	t.deepEqual(queryString.parse('foo=0x11&bar=12.00', {parseNumbers: true}), {foo: 17, bar: 12});
+});
+
+test('NaN value returns as string if option is set', t => {
+	t.deepEqual(queryString.parse('foo=null', {parseNumbers: true}), {foo: 'null'});
+	t.deepEqual(queryString.parse('foo=100a&bar=100', {parseNumbers: true}), {foo: '100a', bar: 100});
+});
```
sindresorhus: Can you include tests that should return `null` and `undefined` as values?

sindresorhus:
Pretty close to being done. The merge conflict needs to be resolved.

sindresorhus:
Looks good. Thanks :)

PR #172, opened 2019-03-27
Fix `split-on-first` dependency

talbenmoshe:
Build is failing as split-on-first is devDependancy and not a "regular" dependancy


sindresorhus:
Ugh, I totally missed this.

talbenmoshe:
No worries.. Thank you!

PR #173, opened 2019-04-09
Support stringifying array containing value 0

padupuy:
For whatever reason, I need to send these kind of query strings to a server:
```
idList=0,1,2,3
```
Before : 0 value was ignored.  
After : 0 value is well added to the final string


padupuy:
Could you review please @sindresorhus @kevva @SamVerschueren  ?

PR #177, opened 2019-05-31
Add `sort` option to `.parse()`

### LINKED ISSUE #176: Query param ordering is changed on .parse
For example this:
```
?all=monkeys&-all=frank&make=canon&page=1&resultSize=20
```
becomes this:
```
{-all: "frank", all: "monkeys", make: "canon", page: "1", resultSize: "20"}
```

Expected behaviour is that it would produce this:
```
{all: "monkeys", -all: "frank", make: "canon", page: "1", resultSize: "20"}
```

It looks as though it's being sorted on purpose:
https://github.com/sindresorhus/query-string/blob/master/index.js#L205

I noticed that you have an option to disable sorting on `stringify`. Is it possible to have the same option on `parse`?

ndtreviv:
In this version we introduce a parse function option called `sort` that defaults to `true` (current default behaviour). When `sort` is `false`, query string keys are not sorted when parsed, and appear in the resulting object in the order they appear in the query string.

Fixes #176 
Closes #162

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,11 +36,11 @@ export interface ParseOptions {
 	readonly arrayFormat?: 'bracket' | 'index' | 'comma' | 'none';
 
 	/**
-	 * @default `false`
-	 * 
 	 * Supports both `Function` as a custom sorting function or `false` to disable sorting.
 	 *
 	 * If omitted, keys are sorted using `Array#sort`, which means, converting them to strings and comparing strings in Unicode code point order.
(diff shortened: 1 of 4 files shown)


ndtreviv:
Ah nuts! I've just realised that this PR is the same as #162 !
Sorry for the duplicate. On the upside...it looks like I've done all the things you were asking for on the other PR that are holding it up 👍 Take your pick!

PR #179, opened 2019-06-04
Fix TypeScript 3.5 compatibility

### LINKED ISSUE #180: does not define the type of the object in the stringify function
I have a problem with typescript 3.5.1. If the version is lower then everything is fine.

I have an object with the interface:
```
interface SuperData {
  first: 'one' | 'two' | 'three';
  second?: boolean | string;
  third?: boolean | string;
  fourth: number[];
  fifth: string[];
  sixth: string[];
}
```

And when I use it:
```
const newSuperData = queryString.stringify(superData)
```

I get an error like:
```
Argument of type 'SuperData' is not assignable to parameter of type '{ [key: string]: unknown; }'. Index signature is missing in type 'SuperData'.
```

I fixed the error here #179

viczhuravlev:
Fixes #180

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -132,7 +132,7 @@ export interface StringifyOptions {
 /**
  * Stringify an object into a query string and sorting the keys.
  */
-export function stringify<T = {[key: string]: unknown}>(
+export function stringify<T extends { [key: string]: any }>(
 	object: T,
 	options?: StringifyOptions
 ): string;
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don’t really see the point of making it generic, but I guess we have to switch from `unknown` to `any`, per https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/

viczhuravlev:
I agree. Updated.

PR #181, opened 2019-06-10
Fix readme examples

laat:
formatting parse result to encode information about keys
```
//=> {foo: ['1', '2', '3']}
//=> {'foo[]': [1, 2, 3]}
```

Fixed some other small issues

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -109,8 +109,8 @@ Type: `boolean`<br>
 Default: `false`
 
 ```js
-queryString.parse('foo[]=1&foo[]=2&foo[]=3', {parseNumbers: true});
-//=> {'foo[]': [1, 2, 3]}
+queryString.parse('foo=1', {parseNumbers: true});
+//=> {foo: 1}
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -110,7 +110,7 @@ Default: `false`
 
 ```js
 queryString.parse('foo[]=1&foo[]=2&foo[]=3', {parseNumbers: true});
-//=> foo: [1, 2, 3]
+//=> {'foo[]': [1, 2, 3]}
```
laat: behaves like this today, is this expected?

sindresorhus: I think we need a simpler example here. Maybe just `queryString.parse('foo=1', {parseNumbers: true});`

sindresorhus:
Thanks for catching these mistakes. Can you also update the examples in https://github.com/sindresorhus/query-string/blob/master/index.d.ts ?

laat:
will do

PR #183, opened 2019-06-11
Add `parseBooleans` option to `.parse()`

### LINKED ISSUE #182: Add parseBooleans option
Parsing an object like 
```javascript
{
    test: true
}
```
results in `test=true`. Using stringify it results in
```javascript
{
    test: "true"
}
```

That's why I think we need an additional option like parseNumbers for booleans.

felixoi:
Fixes #182

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -76,16 +76,16 @@ export interface ParseOptions {
 	readonly parseNumbers?: boolean;
 
 	/**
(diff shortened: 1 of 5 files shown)


sindresorhus:
Note that I cannot release this until https://github.com/sindresorhus/query-string/issues/184 is resolved.

PR #185, opened 2019-06-19
Fix the TypeScript types

### LINKED ISSUE #184: broken types for `parse`
by adding `|number` to the value side of ParsedQuery, you've changed the return type in a really breaking the type signature for users who don't use `parseNumbers` - now every use of this generates errors like this:
```
[devserver] TS2345: Argument of type 'string | number | ReactText[]' is not assignable to parameter of type 'string | string[]'.
[devserver]   Type 'number' is not assignable to type 'string | string[]'.
```
This is only going to get worse when #183 is fixed.

I'd like to suggest either a complex typescript signature such that only passing `parseNumbers` results in the new `string |number| (number|string)[]`, or introduce a new  function `parseValues` which does the number/boolean/etc

cdeutsch:
We need to return different versions of `ParsedQuery` based on the values of `parseBooleans` and `parseNumbers`.

Not sure about the naming. A bit verbose, but....

Fixes #184

@alecf @felixoi will this work for you?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -90,7 +90,7 @@ export interface ParseOptions {
 }
 
 export interface ParsedQuery<T = string> {
-	[key: string]: T | Array<T> | null | undefined;
+	[key: string]: T | T[] | null | undefined;
 }
 
 /**
--- a/index.d.ts
+++ b/index.d.ts
@@ -112,9 +112,9 @@ The returned object is created with [`Object.create(null)`](https://developer.mo
 
 @param query - The query string to parse.
 */
-export function parse(query: string, options: { parseBooleans: true, parseNumbers: true } & ParseOptions): ParsedQueryWithBooleansAndNumbers;
-export function parse(query: string, options: { parseBooleans: true } & ParseOptions): ParsedQueryWithBooleans;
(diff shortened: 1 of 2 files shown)


cdeutsch:
Note I also removed the `readonly` from `ParsedQuery` and its variants, because `parse` is not returning a readonly object. You are able to modify its results (and I do in my code). Maybe I'm missing something there?

index.d.ts
```diff
@@ -90,7 +90,19 @@ export interface ParseOptions {
 }
 
 export interface ParsedQuery {
-	readonly [key: string]: string | number | boolean | Array<string | number | boolean> | null | undefined;
+	[key: string]: string | Array<string> | null | undefined;
```
alecf: you could possibly simplify this as

```typescript
export interface ParseQuery<T> {
  [key: string]: T | Array<T> | null | undefined;
}
```

and then you can just use:

```typescript
export function parse(query: string, options: { parseBooleans: true, parseNumbers: true } & ParseOptions): ParsedQuery<string | number | boolean>;
```

etc..

cdeutsch: Good call. I'll change it to that

cdeutsch: Updated. Thanks! 👍

cdeutsch:
Oops. Accidentally deleted the branch from my repo when cleaning up.

PR #187, opened 2019-06-21
Fix using the `parseNumbers` and `parseBooleans` options together

### LINKED ISSUE #186: TypeError when both parseBooleans and parseNumbers are enabled
Repro:
```js
const queryString = require('query-string');

queryString.parse(
  'num=1.23&bool=false', { parseNumbers: true, parseBooleans: true }
);
```
Results in error
```
/home/user/project/node_modules/query-string/index.js:209
                if (options.parseBooleans && value !== null && (value.toLowerCase() === 'true' || value.toLowerCase() === 'false')) {
                                                                      ^

TypeError: value.toLowerCase is not a function
```

Using query-string 6.8.0 and node 10.16.0

felixoi:
Not tested but I expect it to work. Problem was that if it's was a number, value was set to number already and it still checked if it's a boolean.

Fixes #186

### DIFF
--- a/index.js
+++ b/index.js
@@ -204,9 +204,7 @@ function parse(input, options) {
 
 		if (options.parseNumbers && !Number.isNaN(Number(value))) {
 			value = Number(value);
-		}
-
-		if (options.parseBooleans && value !== null && (value.toLowerCase() === 'true' || value.toLowerCase() === 'false')) {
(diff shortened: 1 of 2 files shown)


felixoi:
Added test so I can confirm it's working @sindresorhus

PR #197, opened 2019-07-29
Add option to skip null and undefined values in stringify

### LINKED ISSUE #196: `skipNulls` option for queryString.stringify
Hi!

```js

const obj1 = {
  a: 1,
  b: '2',
  c: [3, 4, 5],
  d: true,
  e: 'just string',
  f: null,
  g: undefined,
};

// qs
  addQueryPrefix: true,
  arrayFormat: 'comma',
  skipNulls: true,
});
const ex3 = qs.parse(ex2, {
  parseArrays: true,
  comma: true,
  ignoreQueryPrefix: true,
});

// query-string
  strict: true,
});
  parseNumbers: true,
  parseBooleans: true,
});

console.log('ex2', ex2);
console.log('ex3', ex3);
console.log('ex4', ex4);
console.log('ex5', ex5);
```

FredrikAugust:
You can now pass an option called `skipNullAndUndefined` to the options
object in stringify to prevent the generator from generating keys for
those keys.

Fixes #196.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -100,9 +100,9 @@ The returned object is created with [`Object.create(null)`](https://developer.mo
 
 @param query - The query string to parse.
 */
-export function parse(query: string, options: {parseBooleans: true, parseNumbers: true} & ParseOptions): ParsedQuery<string | boolean | number>;
-export function parse(query: string, options: {parseBooleans: true} & ParseOptions): ParsedQuery<string | boolean>;
-export function parse(query: string, options: {parseNumbers: true} & ParseOptions): ParsedQuery<string | number>;
+export function parse(query: string, options: { parseBooleans: true, parseNumbers: true } & ParseOptions): ParsedQuery<string | boolean | number>;
+export function parse(query: string, options: { parseBooleans: true } & ParseOptions): ParsedQuery<string | boolean>;
+export function parse(query: string, options: { parseNumbers: true } & ParseOptions): ParsedQuery<string | number>;
 export function parse(query: string, options?: ParseOptions): ParsedQuery;
 
 export interface ParsedUrl {
@@ -192,13 +192,33 @@ export interface StringifyOptions {
 	```
 	*/
 	readonly sort?: ((itemLeft: string, itemRight: string) => number) | false;
+
+	/**
+	Takes a boolean which decides whether or not `undefined` and `null` values should be skipped when stringify-ing.
+
+	@default false
+
+	@example
+	```
+	queryString.stringify({ a: 'ja', b: undefined, c: null, d: 'vielsker' }, {
+		skipNullAndUndefined: true,
+	});
+	// => 'a=ja&d=vielsker'
+
+	queryString.stringify({ a: undefined, b: null }, {
+		skipNullAndUndefined: true,
+	});
+	// => ''
+	```
+	*/
+	readonly skipNullAndUndefined?: boolean;
 }
 
 /**
 Stringify an object into a query string and sort the keys.
 */
 export function stringify(
-	object: {[key: string]: any},
+	object: { [key: string]: any },
 	options?: StringifyOptions
 ): string;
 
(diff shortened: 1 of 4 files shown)


FredrikAugust:
Appears that `Object.entries` is not supported in older versions of Node.js 6. Will get to this now.

index.js
```diff
@@ -243,6 +243,15 @@ exports.stringify = (object, options) => {
 	}, options);
 
 	const formatter = encoderForArrayFormat(options);
+
+	if (options.skipNullAndUndefined) {
+		for (const key of Object.keys(object)) {
+			if (object[key] === undefined || object[key] === null) {
+				delete object[key];
```
goodwin64: Here you're mutating the passed object.

goodwin64: I suppose `stringify` function has to be pure.

FredrikAugust: Makes sense. Will update it now.

FredrikAugust: Fixed in #d0b21db.

sindresorhus:
@yaodingyd Can you help review? 🙏

index.d.ts
```diff
@@ -100,9 +100,9 @@ The returned object is created with [`Object.create(null)`](https://developer.mo
 
 @param query - The query string to parse.
 */
-export function parse(query: string, options: {parseBooleans: true, parseNumbers: true} & ParseOptions): ParsedQuery<string | boolean | number>;
-export function parse(query: string, options: {parseBooleans: true} & ParseOptions): ParsedQuery<string | boolean>;
-export function parse(query: string, options: {parseNumbers: true} & ParseOptions): ParsedQuery<string | number>;
+export function parse(query: string, options: { parseBooleans: true, parseNumbers: true } & ParseOptions): ParsedQuery<string | boolean | number>;
+export function parse(query: string, options: { parseBooleans: true } & ParseOptions): ParsedQuery<string | boolean>;
+export function parse(query: string, options: { parseNumbers: true } & ParseOptions): ParsedQuery<string | number>;
```
sindresorhus: Don't do unrelated changes. Same in many other places.

index.d.ts
```diff
@@ -192,13 +192,33 @@ export interface StringifyOptions {
 	```
 	*/
 	readonly sort?: ((itemLeft: string, itemRight: string) => number) | false;
+
+	/**
+	Takes a boolean which decides whether or not `undefined` and `null` values should be skipped when stringify-ing.
```
sindresorhus: This could be written better. Try to make it more concise.

index.js
```diff
@@ -243,6 +243,17 @@ exports.stringify = (object, options) => {
 	}, options);
 
 	const formatter = encoderForArrayFormat(options);
+
+	const object = Object.assign({}, input);
+
+	if (options.skipNullAndUndefined) {
+		for (const key of Object.keys(object)) {
+			if (object[key] === undefined || object[key] === null) {
```
yaodingyd: It doesn't work with `arrayFormat`. You need to check if `object[key]` is an array, and iterate through array to check null/undefined.

test/stringify.js
```diff
@@ -193,3 +193,32 @@ test('should disable sorting', t => {
...
+});
+
+test('should ignore both null and undefined when skipNullAndUndefined is set', t => {
+	t.is(queryString.stringify({
+		a: undefined,
+		b: null
+	}, {
+		skipNullAndUndefined: true
+	}), '');
+});
```
yaodingyd: Add test cases for `arrayFormat`.

MichaelDeBoey: Maybe also add test cases for when it's set to `false`? 🤔
'Cause I thought this was already the current behavior.

sindresorhus:
Also note https://github.com/sindresorhus/query-string/issues/196#issuecomment-522169947

sindresorhus:
@FredrikAugust Still interested in finishing this?

FredrikAugust:
It would be nice if someone else could take over, don't have much spare time right now to work on this.

yaodingyd:
I can pick this up over the weekend, if it's ok.

PR #198, opened 2019-08-05
Bug: Raises hard error in older versions of Firefox

kylemellander:
In older browsers (I found it broken in Firefox 48.0.2), this would raise the following error: 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Missing_initializer_in_const

Using `let` instead prevents this error since the variable is overwritten each iteration.

### DIFF
--- a/index.js
+++ b/index.js
@@ -195,7 +195,7 @@ function parse(input, options) {
 		return ret;
 	}
 
-	for (const param of input.split('&')) {
+	for (let param of input.split('&')) {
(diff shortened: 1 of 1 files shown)


sindresorhus:
You need to transpile with Babel if you need to support older Firefox versions (and other older browsers).

You can find a more detailed explanation here: https://github.com/sindresorhus/ama/issues/446

If you use Create React App, upgrade to [Create React App v2](https://reactjs.org/blog/2018/10/01/create-react-app-v2.html). It supports automatic transpilation of dependencies, which will make this package *just work*.

PR #202, opened 2019-08-08
Make `parseNumbers` and `parseBooleans` options work with `arrayFormat`

### LINKED ISSUE #199: parseNumbers ignored with arrayFormat
The `parseNumbers` option does not get applied to values in an array.  e.g.:

```
const query = queryString.parse('?arrayOfNumbers=15,16,17&singleNumber=18', {
  parseNumbers: true,
  arrayFormat: 'comma',
});
```

Produces:

```
{
  arrayOfNumbers: ["15", "16", "17"],
  singleNumber: 18,
}
```

I was expecting it to produce:

```
{
  arrayOfNumbers: [15, 16, 17],
  singleNumber: 18,
}
```

yaodingyd:
fix #199

### DIFF
--- a/index.js
+++ b/index.js
@@ -171,6 +171,16 @@ function extract(input) {
 	return input.slice(queryStart + 1);
 }
 
+function parseValue(value, options) {
+	if (options.parseNumbers && !Number.isNaN(Number(value)) && (typeof value === 'string' && value.trim() !== '')) {
+		value = Number(value);
+	} else if (options.parseBooleans && value !== null && (value.toLowerCase() === 'true' || value.toLowerCase() === 'false')) {
+		value = value.toLowerCase() === 'true';
+	}
+
+	return value;
+}
+
 function parse(input, options) {
 	options = Object.assign({
 		decode: true,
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -201,16 +211,20 @@ function parse(input, options) {
...
-		if (options.parseNumbers && !Number.isNaN(Number(value)) && (typeof value === 'string' && value.trim() !== '')) {
-			value = Number(value);
-		} else if (options.parseBooleans && value !== null && (value.toLowerCase() === 'true' || value.toLowerCase() === 'false')) {
-			value = value.toLowerCase() === 'true';
-		}
-
 		formatter(decode(key, options), value, ret);
 	}
 
+	Object.keys(ret).forEach(key => {
```
sindresorhus: Use a `for-of` loop. Same with line 220.

yaodingyd: Use `for-of` would only get me `value`s, but I need `key`s to reassign `ret`

index.js
```diff
@@ -201,16 +211,20 @@ function parse(input, options) {
...
+		const value = ret[key];
+		if (Boolean(value) && typeof value === 'object') {
```
sindresorhus: Shouldn't this test for array?

index.js
yaodingyd: Also `for-of` can't be used with objects, right?

index.js
yaodingyd: If `arrayFormat` is set to `index`, ret would be an object. I want to make this function generic so it can work both on array and object.

index.js
sindresorhus:
```js
for (const key of Object.keys(ret)) {
```

index.js
sindresorhus: `Boolean(value) &&` can be removed. And the check should be:

```suggestion
		if (typeof value === 'object' && value !== null) {
```

mikestecker:
Awesome, this solves an issue I just came across. Can you release a new version with this soon?

sindresorhus:
Done

PR #213, opened 2019-10-02
Added option to skip nulls, undefined, and empty strings for stringify

### LINKED ISSUE #196: `skipNulls` option for queryString.stringify

CMilby:
Added optional parameters to stringify

- skipUndefined: Skip or keep values that are undefined (Original behavior is default)
- skipNulls: Allows to skip or keep null values
- skipNullAndUndefined: Behavior of skipUndefined and skipNulls
- skipEmptyStrings: Skips values with empty strings

Fixes #196

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -192,6 +192,64 @@ export interface StringifyOptions {
 	```
 	*/
 	readonly sort?: ((itemLeft: string, itemRight: string) => number) | false;
+
+	/**
+	Takes a boolean which decides if 'undefined' values should be skipped during stringify
+
+	@default true
+
+	@example
+	```
+	queryString.stringify({a: undefined, b: null, c: 'one'}, {skipUndefined=true});
+	//=> 'b&c=one'
+	```
+	*/
+	readonly skipUndefined?: boolean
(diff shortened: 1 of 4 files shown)


sindresorhus:
We've already added a `skipNull` option: https://github.com/sindresorhus/query-string#skipnull I don't really see the need to change the default behavior for `undefined` or filter empty strings. Can you provide some real-world use-cases for the additional options?

PR #214, opened 2019-10-03
Only parse `+` to space when `decode` is true

### LINKED ISSUE #119: + is parsed as white space
I see there is a test that indicated `+` should be parsed into white space, however when I test this here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
`+` is encoded as `%2B`.

Anusien:
The parsing of '+' to ' ' is a constant source of confusion in the query-string module. This is correct behavior, but I think it should only occur when decode is set to true. This is because this decoding isn't handled by the `decode-uri-component`, but it's being done manually in this project. See [this comment](https://github.com/sindresorhus/query-string/issues/119#issuecomment-533640064).

Fixes #119

### DIFF
--- a/index.js
+++ b/index.js
@@ -206,7 +206,7 @@ function parse(input, options) {
 	}
 
 	for (const param of input.split('&')) {
-		let [key, value] = splitOnFirst(param.replace(/\+/g, ' '), '=');
+		let [key, value] = splitOnFirst(options.decode ? param.replace(/\+/g, ' ') : param, '=');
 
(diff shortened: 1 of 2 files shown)


Anusien:
@sindresorhus what do you think?

sindresorhus:
Shouldn't `encode` be changed too then?

Anusien:
I don't see where `encode` does any replacement manually.

PR #215, opened 2019-10-06
Add option to skip keys with the value of `null` in `.stringify()`

### LINKED ISSUE #196: `skipNulls` option for queryString.stringify

yaodingyd:
Follow up on #197 to fix #196.

Closes #197

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -195,13 +195,16 @@ export interface StringifyOptions {
 
 	/**
 	Skip the key in stringify result if value is `null` or `undefined`.
+
 	@default false
+
 	@example
 	```
 	queryString.stringify({a: 1, b: undefined, c: null, d: 4}, {
 		skipNullAndUndefined: true,
 	});
 	//=> 'a=1&d=4'
+
 	queryString.stringify({a: undefined, b: null}, {
 		skipNullAndUndefined: true,
 	});
--- a/index.d.ts
+++ b/index.d.ts
@@ -224,5 +224,7 @@ export function stringify(
 
 /**
 Extract a query string from a URL that can be passed into `.parse()`.
+
+Note: This behaviour can be changed with the `skipNullAndUndefined` option.
 */
 export function extract(url: string): string;
--- a/index.d.ts
+++ b/index.d.ts
@@ -201,17 +201,17 @@ export interface StringifyOptions {
 	@example
 	```
 	queryString.stringify({a: 1, b: undefined, c: null, d: 4}, {
-		skipNullAndUndefined: true,
+		skipNull: true,
 	});
 	//=> 'a=1&d=4'
 
 	queryString.stringify({a: undefined, b: null}, {
-		skipNullAndUndefined: true,
+		skipNull: true,
 	});
 	//=> ''
 	```
 	*/
-	readonly skipNullAndUndefined?: boolean;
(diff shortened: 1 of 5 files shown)


yaodingyd:
Additionally, fix a bug in current stringify behavior: if `arrayFormat` is set to `comma` and first item is null, the result won't have the key in it. something like
```javascript
queryString.stringify({a: [null, 1, undefined, 2, null]}, {
		arrayFormat: 'comma'
	}
// => `,1,2`, missing `a=`
```

sindresorhus:
What about https://github.com/sindresorhus/query-string/issues/196#issuecomment-522169947 ?

index.js
```diff
@@ -257,7 +257,17 @@ exports.stringify = (object, options) => {
 	}, options);
 
 	const formatter = encoderForArrayFormat(options);
-	const keys = Object.keys(object);
+
+	const objectCpy = Object.assign({}, object);
```
sindresorhus:
```suggestion
	const objectCopy = Object.assign({}, object);
```

readme.md
```diff
@@ -204,10 +204,33 @@ queryString.stringify({b: 1, c: 2, a: 3}, {sort: false});
 
 If omitted, keys are sorted using `Array#sort()`, which means, converting them to strings and comparing strings in Unicode code point order.
 
+##### skipNullAndUndefined
+
+Skip the key in stringify result if value is `null` or `undefined`.
```
sindresorhus:
```suggestion
Skip the element in the stringify result if the value is `null` or `undefined`.
```

readme.md
```diff
@@ -204,10 +204,33 @@ queryString.stringify({b: 1, c: 2, a: 3}, {sort: false});
...
+	skipNullAndUndefined: true,
+});
+//=> ''
+```
+
 ### .extract(string)
 
 Extract a query string from a URL that can be passed into `.parse()`.
 
+Note: This behaviour can be changed with the `skipNullAndUndefined` option.
```
sindresorhus: This should be added to index.d.ts too.

yaodingyd:
which direction are we going? the option should be called `skipNull` or change the current behavior so that `undefined` is not skipped?

sindresorhus:
> the option should be called skipNull 

This

index.d.ts
```diff
@@ -192,6 +192,26 @@ export interface StringifyOptions {
 	```
 	*/
 	readonly sort?: ((itemLeft: string, itemRight: string) => number) | false;
+
+	/**
+	Skip the key in stringify result if value is `null` or `undefined`.
```
sindresorhus: This makes it sound like it will **not** skip `undefined` by default, which it does.

Issue #231, opened 2020-01-06
Encode empty arrays


mitar:
For my use case it would be great if there would be a way to encode an empty array as well. For example, in `bracket` style, maybe `{a: []}` could be encoded as `a[]` only (while `{a: [1]}` is encoded as `a[]=1`).

sindresorhus:
See my answer in https://github.com/sindresorhus/query-string/issues/228#issuecomment-575197530. It applies to this too.

mitar:
Yes, but `bracket` style does not help here (which is your suggestion there).

So `bracket` style is less ambiguous, so adding `a[]` would solve this only ambiguous I know of which still exists for this style.

sindresorhus:
That style is already used for null values: https://github.com/sindresorhus/query-string/blob/db66e68fc7e323995de713c38504a192472038fa/test/stringify.js#L117

mitar:
I think `bar[]=` should be an empty array, while `bar` should be `null`, `bar=null` is string `null`, and `bar=` is empty string.

sindresorhus:
Closing as implemented. Use `arrayFormat: 'bracket-separator'` (optionally with `arrayFormatSeparator`) to represent empty arrays.
> `stringify({a: []}) → 'a[]'` and `parse('a[]') → {a: []}`.
> `a[]=` remains an array with a single empty string, so the cases are unambiguous and round-trip.

mitar:
Awesome, thanks!

Issue #210, opened 2019-09-11
Support predefining types


goodwin64:
(related to https://github.com/sindresorhus/query-string/issues/201)

## Consider 2 examples:

```ts
const search1 = '?phoneNumber=%2B380951234567&subtopicId=2&topicId=1';
const result1 = queryString.parse(search, {
  parseBooleans: true,
  parseNumbers: true,
});

const search2 = '?phoneNumber=a%2B380951234567&subtopicId=2&topicId=1';
const result2 = queryString.parse(search, {
  parseBooleans: true,
  parseNumbers: true,
});

console.log('result1', result1);
// {
//   phoneNumber: 380951234567,
//   subtopicId: 2,
//   topicId: 1,
// };
console.log('result2', result2);
// {
//   phoneNumber: 'a+380951234567',
//   subtopicId: 2,
//   topicId: 1,
// };
```

In the 1st example, "+" sign (`%2B` escaped) is ignored thus reducing to the number parsing, which is not what it supposed to be.

The main goal is to keep parsing numbers for parameters which values **starts** on the number and parse as plain string **otherwise**.

Does it look feasible to make a PR on this or should I not use `parseNumbers` option in that case?

sindresorhus:
`+380951234567` is a valid number though. You would expect `'-380951234567'` to be parsed as a number, right? That being said, I agree it's not a optimal behavior.

// @yaodingyd

yaodingyd:
What should be the optimal behavior? Here it's using Number() function to convert to number.

goodwin64:
I meant if I store the number "+123" in the URL and then I parse it with "parseNumbers" option I get lost the plus sign (+) and the parsed value will be just "123".

Meanwhile I want to store some number-like parameters in URL so removing the "parseNumbers" option is not the right solution for me.

goodwin64:
So when I prepend the value with the prefix "a" it's just a workaround not to lost the plus sign :)

"a+123" get parsed as is and then I just skip the "a" letter

sindresorhus:
I think some kinda schema support here could solve the ambiguity. You could then explicitly define the type for this one key.  See: https://github.com/sindresorhus/query-string/issues/228#issuecomment-575197530

iagolaguna:
Hi, I would like to help you to solve this issue, I was thinking about you said about schema.

What do you think about something like this

```
const search2 = '?phoneNumber=a%2B380951234567&subtopicId=2&topicId=1';
[...]

PR #238, opened 2020-02-04
add allowEmptyArrays option

stevenmiller888:
This is one possible solution to #231. Solves it by adding an `allowEmptyArrays` option, which will do the following, for example:

stringify: `{ a: [] }` => `'a[]='` (instead of `''`)
parse: `'a[]='` => `{ a: [] }` (instead of `{ a: [null] }`).

If we like the idea, I'm happy to clean it up and add docs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -96,7 +96,12 @@ function parserForArrayFormat(options) {
 				}
 
 				if (accumulator[key] === undefined) {
-					accumulator[key] = [value];
+					if (options.allowEmptyArrays) {
+						accumulator[key] = [];
+					} else {
+						accumulator[key] = [value];
+					}
+
 					return;
 				}
 
@@ -197,7 +202,8 @@ function parse(input, options) {
 		sort: true,
 		arrayFormat: 'none',
 		parseNumbers: false,
-		parseBooleans: false
+		parseBooleans: false,
+		allowEmptyArrays: false
 	}, options);
 
 	const formatter = parserForArrayFormat(options);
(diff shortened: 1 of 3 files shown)


sindresorhus:
This would indeed solve this exact problem, but there are many other problems with arrays. I think it would be better to solve it for real by letting to explicitly specify that it's an array. Then there's no possible ambiguity. See: https://github.com/sindresorhus/query-string/issues/210#issuecomment-586300701 Thoughts on that?

// @Mitar

mitar:
@stevenmiller888 I like it, but how do we encode null then now?

mjsarfatti:
This has been causing unexpected results and bugs in my applications. Bugs extremely hard to detect, especially since this library is used by other libraries in turn.

Just spent half a day pinpointing the issues, and I finally realised the problem is in query-string.

If in a form you have a set of checkboxes, you can send it to the server as an array of params.
Now, if the user deselects all the checkboxes, how do I tell the server that the selection is now empty? I would assume that if I send a param as an array, empty or not, this param will reach the server. 

In my opinion this shouldn't even be behind a flag, but it should be standard behaviour. Think about it. If you were writing an application that can send an array of values to the server, would you ever think that sending an empty array would have no effect and would actually send nothing?

If making it standard we are breaking past behaviours, maybe it should be a new major release. If that is not possible at the moment, would you be able to release it behind the flag as @stevenmiller888 suggested? This has been open for some time now.

This is just my opinion anyway, all my gratitude for the work you put towards this package 🙏

DV8FromTheWorld:
Instead of a flag to allow empty arrays in other formats, it might be better to introduce a format that is more explicit. 
Check out the proposal for `bracket-separator` and the question in regard to handling empty arrays https://github.com/sindresorhus/query-string/pull/276

PR #241, opened 2020-02-18
Add a condition to check and make sure the query string parsed to a n…

SaidShah:
…umber is not solved as an equation when it contains a + or an e or any other math characters

When the values in the query string are converted to numbers, if they contain values such as +,-,*,e they will be treated as an equation and solved. for example if the value being parsed is 192e11 is converted to 192000000 and 1920000000000000000000000 gets converted to 1.92e+24 etc..

### DIFF
--- a/index.js
+++ b/index.js
@@ -190,10 +190,15 @@ function extract(input) {
 }
 
 function parseValue(value, options) {
+	const initialValue = value;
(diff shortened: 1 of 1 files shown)


sindresorhus:
This will needs tests. I think the behavior should be made explicit in the readme. This is also potentially breaking if anyone is relying on the current behavior.

sindresorhus:
Bump

sindresorhus:
https://github.com/sindresorhus/query-string/issues/263

Issue #242, opened 2020-02-20
Issue for ISO 8601 dates on arrayFormat separator option


jessedobbelaere:
On v6.10.1 we were succesfully parsing the following two url params using the `comma` parser:

```javascript
parse(queryString, { arrayFormat: "comma", parseBooleans: true, parseNumbers: true });
// Before: id%28in%29=992232%2C992237%2C992240
// After: 
// {
//    id(in): [992232, 992237, 992240],
// }

// Before: timestamp_begin(gte)=2020-01-01T00:00:00%2B03:00
// After: 
// {
//     timestamp_begin(gte): "2020-01-01T00:00:00+03:00"
// }
```

Here, our ISO 8601 datetime is not getting modified ✅ .

However, with the release of v6.11.0, we ran into issues because of the encoded commas that are passed to our API, due to the change of (#236). 
After switching to the arrayFormat "separator" we seem to get the old behaviour from v6.10.1 with our encoded comma arrays, except for one issue with our ISO 8601 DateTime.

```javascript
parse(queryString, { arrayFormat: "separator", parseBooleans: true, parseNumbers: true });
// Before: id%28in%29=992232%2C992237%2C992240
// After: 
// {
//    id(in): [992232, 992237, 992240],
// }
// This works great again

// Before: timestamp_begin(gte)=2020-01-01T00:00:00%2B03:00
// After: 
// {
//     timestamp_begin(gte): "2020-01-01T00:00:00 03:00"
// }
// Here the + sign is removed from our datetime string
```

Note: when switching it to `arrayFormat: "comma"` again the datetime is correct (+ sign not removed. (But we cannot use the comma format because of our encoded comma's in the requests we get.)

Made a small code sample to illustrate: https://stackblitz.com/edit/typescript-udk9sb (open console)

sindresorhus:
// @cheqianxiao

jessedobbelaere:
hi @cheqianxiao, do you know if there will be a fix in the future? Or if there is a workaround for:

https://github.com/sindresorhus/query-string/pull/243#issuecomment-591179715

As the PR fixed the issue partially

cheqianxiao:
@jessedobbelaere  hey~  you can try  this workround for that issue:
```javascript
parse(decodeURIComponent(queryString), { arrayFormat: "comma", parseBooleans: true, parseNumbers: true });
 ```
and...  sorry, I don't know if there will be a fix in the future.

PR #243, opened 2020-02-25
Fix value being decoded twice with `arrayFormat` option set to `separator`

cheqianxiao:
fix part of  #242  ( caused by #236

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -311,13 +311,15 @@ test('query strings having comma encoded and format option as `comma`', t => {
 	});
 });
 
-test('value shoud not be decoded twice with arrayFormat set as `separator`', t => {
+test('value should not be decoded twice with arrayFormat set as `separator`', t => {
 	t.deepEqual(queryString.parse('foo=2020-01-01T00:00:00%2B03:00', {arrayFormat: 'separator'}), {
 		foo: '2020-01-01T00:00:00+03:00'
 	});
 });
 
 // See https://github.com/sindresorhus/query-string/pull/243#issuecomment-591179715
-test.failing('value separated by encoded comma will not be parsed as array with arrayFormat set as `comma`', t => {
-	t.fail();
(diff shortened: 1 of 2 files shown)


sindresorhus:
> part of 242

What part does this not fix?

sindresorhus:
You can use Array#includes instead of indexOf

cheqianxiao:
> > part of 242
> 
> What part does this not fix?

it not fix : 
```js
parse(queryString, { arrayFormat: "comma", parseBooleans: true, parseNumbers: true });
// Request: id%28in%29=992232%2C992237%2C992240
// Parsed: 
// {
//    id(in): '992232,992237,992240'
// }
```

sindresorhus:
Can you add a [failing test](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests) for that case that you're not fixing?

test/parse.js
```diff
@@ -310,3 +310,14 @@ test('query strings having comma encoded and format option as `comma`', t => {
...
+
+test('value shoud not be decoded twice with arrayFormat set as `separator`', t => {
+	t.deepEqual(queryString.parse('foo=2020-01-01T00:00:00%2B03:00', {arrayFormat: 'separator'}), {
+		foo: '2020-01-01T00:00:00+03:00'
+	});
+});
+
+// See https://github.com/sindresorhus/query-string/pull/243#issuecomment-591179715
+test.failing('value separated by encoded comma will not be parsed as array with arrayFormat set as `comma`', t => {
+	t.fail();
```
sindresorhus, line 322: This should actually be a valid test implementation, which would pass when the issue is fixed.

test/parse.js
```diff
@@ -310,3 +310,14 @@ test('query strings having comma encoded and format option as `comma`', t => {
 		]
 	});
 });
+
+test('value shoud not be decoded twice with arrayFormat set as `separator`', t => {
```
sindresorhus, line 314: Typo

PR #252, opened 2020-03-21
Add `skipEmptyString` option

### LINKED ISSUE #244: Add parameter skipEmptyString
I love the skipNull that is available in the stringify.
Would it be posible to also include a skipEmpty parameter that filters all params where the value = ""

mum-never-proud:
Fixes #244

### DIFF
--- a/index.js
+++ b/index.js
@@ -269,10 +269,6 @@ function parse(input, options) {
 	}, Object.create(null));
 }
 
-function isEmptyString(value) {
-	return value === '';
-}
-
 function isNullOrUndefined(value) {
 	return value === null || value === undefined;
 }
@@ -300,7 +296,7 @@ exports.stringify = (object, options) => {
 		.keys(object)
 		.reduce((objectCopy, key) => {
 			if ((options.skipNull && isNullOrUndefined(object[key])) ||
-				(options.skipEmptyString && isEmptyString(object[key]))) {
+				(options.skipEmptyString && object[key] === '')) {
 				return objectCopy;
 			}
 
--- a/index.js
+++ b/index.js
@@ -8,7 +8,9 @@ function encoderForArrayFormat(options) {
 		case 'index':
 			return key => (result, value) => {
 				const index = result.length;
-				if (value === undefined || (options.skipNull && value === null)) {
+				if (value === undefined ||
+					(options.skipNull && value === null) ||
+					(options.skipEmptyString && value === '')) {
 					return result;
 				}
 
@@ -24,7 +26,9 @@ function encoderForArrayFormat(options) {
 
 		case 'bracket':
 			return key => (result, value) => {
-				if (value === undefined || (options.skipNull && value === null)) {
+				if (value === undefined ||
+					(options.skipNull && value === null) ||
+					(options.skipEmptyString && value === '')) {
 					return result;
 				}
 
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -263,6 +269,14 @@ function parse(input, options) {
 	}, Object.create(null));
 }
 
+function isEmptyString(value) {
+	return value === '';
+}
```
sindresorhus: This utility has no value over just checking directly.

sindresorhus:
You also need to update index.d.ts. That one and the readme should be in sync. I would also like to see more tests.

index.js
mum-never-proud: done

index.js
```diff
@@ -282,14 +292,18 @@ exports.stringify = (object, options) => {
...
 	const formatter = encoderForArrayFormat(options);
 
-	const objectCopy = Object.assign({}, object);
-	if (options.skipNull) {
-		for (const key of Object.keys(objectCopy)) {
-			if (objectCopy[key] === undefined || objectCopy[key] === null) {
-				delete objectCopy[key];
+	const objectCopy = Object
+		.keys(object)
+		.reduce((objectCopy, key) => {
```
sindresorhus, line 302: Don't do unrelated changes. I prefer this to stay a `for-of` loop instead of reduce, for readability.

readme.md
```diff
@@ -304,6 +304,31 @@ queryString.stringify({a: undefined, b: null}, {
 //=> ''
 ```
 
+##### skipEmptyString
+
+Skip keys with empty strings as the value.
```
sindresorhus, line 309:
```suggestion
Skip keys with an empty string as the value.
```

index.js
mum-never-proud: I didn't refactor, rather a related change

```javascript
if (options.skipNull) {
  for (const key of Object.keys(objectCopy)) {
    if (objectCopy[key] === undefined || objectCopy[key] === null) {
      delete objectCopy[key];
    }
  }
}
```

the same can be achieved with 

```javascript
if (options.skipNull || options. skipEmptyString) {
  for (const key of Object.keys(objectCopy)) {
    if (objectCopy[key] === undefined || objectCopy[key] === null || objectCopy[key] === '') {
      delete objectCopy[key];
    }
  }
}
```
suppose in future if we introduce other filters the list will go on...

is there any better way than this?

sindresorhus: > is there any better way than this?

Do the same as now, but instead of reduce, use `for-of` loop, meaning no `Object.asssign`. You could maybe move the filtering logic here into a helper function (in this scope, not top-level).

mum-never-proud: hmm ok, if I remove the reduce this is how the code is (existing) will this work? sorry if I am not getting your point

```javascript
const objectCopy = Object.assign({}, object);

for (const key of Object.keys(object)) {
    if ((options.skipNull && isNullOrUndefined(object[key])) ||
        (options.skipEmptyString && object[key] === '')) {
        delete objectCopy[key];
    }
}
```

this is another approach

```javascript
const shouldFilter = key => (
    (options.skipNull && isNullOrUndefined(object[key])) ||
    (options.skipEmptyString && object[key] === '')
);

const objectCopy = {};

for (const key of Object.keys(object)) {
    if (!shouldFilter(key)) {
        objectCopy[key] = object[key];
    }
}
```

we can add as many filters we need in future

your thoughts?

sindresorhus: ^ The last code block here is what I had in mind. 👌 Sorry for being unclear.

PR #254, opened 2020-04-23
Add benchmarks

omgovich:
Hi! I love this library and use it on many projects, so I decided to help you with making it better.

I have experience in the size and performance optimizations (see [my password generation library's benchmarks](https://github.com/omgovich/omgopass#benchmark)). I think these things are a super important in the modern frontend world and I believe that every library's goal is to provide the fastest and the smallest (in terms of the bundle size) solution.  So if you don't mind I would help you with that. 

To be able to optimize something, we need tools to measure the size of the library and the speed of each method. So I started with providing these development tools for this repository.

1. I installed [size-limit](https://github.com/ai/size-limit). This library provides the easiest way to calculate gzipped library size including all dependencies. Also, you can install [GitHub action](https://github.com/andresz1/size-limit-action) to automatically check how every PR changes the library's size. 

To check the library size execute `yarn size` in the project root.
```
% yarn size

Size: 1.69 KB with all dependencies, minified and gzipped
```

2. I installed [benchmark.js](https://benchmarkjs.com) and wrote test cases for all library methods. This solution allows us to find places that need optimization and check out how PRs/changes affect performance.  

To run the benchmark execute `yarn benchmark` in the project root.
```
% yarn benchmark

parse_____________________________________83,539 ops/s
parse (parseNumbers=true)_________________82,329 ops/s
parse (parseBooleans=true)________________79,727 ops/s
parse (sort=false)________________________97,193 ops/s
parse (decode=false)_____________________217,655 ops/s
parse (arrayFormat=bracket)_______________67,130 ops/s
parse (arrayFormat=index)_________________65,591 ops/s
parse (arrayFormat=comma)_________________96,522 ops/s
stringify________________________________152,056 ops/s
stringify (strict=false)_________________219,271 ops/s
stringify (encode=false)_________________373,492 ops/s
stringify (skipNull=true)________________155,919 ops/s
stringify (skipEmptyString=true)_________160,193 ops/s
stringify (arrayFormat=bracket)__________149,923 ops/s
stringify (arrayFormat=index)____________134,268 ops/s
stringify (arrayFormat=comma)____________155,819 ops/s
extract_______________________________28,180,181 ops/s
parseUrl__________________________________83,579 ops/s
stringifyUrl_____________________________115,330 ops/s
```

### DIFF
--- a/benchmark.js
+++ b/benchmark.js
@@ -1,5 +1,5 @@
 const Benchmark = require('benchmark');
-const queryString = require('..');
+const queryString = require('.');
 
 const {stringify, stringifyUrl} = queryString;
 const suite = new Benchmark.Suite();
--- a/benchmark.js
+++ b/benchmark.js
@@ -1,3 +1,4 @@
+'use strict';
 const Benchmark = require('benchmark');
 const queryString = require('.');
 
@@ -21,7 +22,7 @@ const TEST_INDEX_STRING = stringify(TEST_OBJECT, {arrayFormat: 'index'});
 const TEST_COMMA_STRING = stringify(TEST_OBJECT, {arrayFormat: 'comma'});
 const TEST_URL = stringifyUrl({url: TEST_HOST, query: TEST_OBJECT});
 
-// Creates test case and adds it to the suite
+// Creates a test case and adds it to the suite
 const defineTestCase = (methodName, input, options) => {
 	const fn = queryString[methodName];
 	const label = options ? ` (${stringify(options)})` : '';
@@ -30,7 +31,8 @@ const defineTestCase = (methodName, input, options) => {
 };
 
 // Define all test cases
-// parse
+
+// Parse
 defineTestCase('parse', TEST_STRING);
 defineTestCase('parse', TEST_STRING, {parseNumbers: true});
 defineTestCase('parse', TEST_STRING, {parseBooleans: true});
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for the PR. I'm happy to take the benchmark, but I'm not interested in using `size-limit`.

Also make sure `$ npm test` passes locally for you.

omgovich:
This test 👇 is crashing since I forked the repo. I think it's broken in `master` branch too.
```
parse › value separated by encoded comma will not be parsed as array with `arrayFormat` option set to `comma`
```

omgovich:
May I ask why you don't like the idea to have `size-limit` installed?
 It's a very popular library made by  Andrey Sitnik (PostCSS and Autoprefixer author).
You don't have to add it to the test flow, but I guess it's good to be able to check the library size and have an easy way compare library size before/after some changes, especially if you decide to add/change some dependency.

P.S. Also I think that it will be helpful for other developers if you add the library's size in the project description. Here is an example. I had `urijs` installed in my project and costs us 30 KB (gzipped), so I decided to check the size of `query-string` and had to go to bundlephobia.com to find out the size of this library. But this site shows the wrong value (the size of the whole NPM package, not the imported code).

sindresorhus:
> May I ask why you don't like the idea to have size-limit installed?

I try very hard to not introduce tooling when it's not important. I've added a simple badge to the readme instead.

sindresorhus:
Thanks :)

PR #261, opened 2020-05-21
Add `.pick()` and `.exclude()` methods

### LINKED ISSUE #259: Feature request: query-string strip (`pick`, `exclude`)
To eliminate unneeded params, preserve only needed
```js
queryString.strip('http://example.com/?a=1&b=2', [ 'a' ] )
// result : http://example.com/?a=1
```

It would be helpful when page has many redudant params (like statistic params) and want to get a cleaner one

---
Better function names from @Uzlopak:
```js
queryString.pick('http://example.com/?a=1&b=2', [ 'a' ] )
// result : http://example.com/?a=1


queryString.exclude('http://example.com/?a=1&b=2', [ 'a' ] )
// result : http://example.com/?b=2
```

BenoitSerrano:
Fixes #259 

Add two new functions `pick` and `exclude` that automatically picks / exclude keys from the query parameters

cc @sindresorhus :)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -329,31 +329,30 @@ export function extract(url: string): string;
 
 
 /**
-	Parses a query string and only picks the passed keysToPicks
+Parses a query string and only picks the passed keysToPicks
 
-	@example
-	```
-	import queryString = require('query-string');
+@example
+```
+import queryString = require('query-string');
 
-	queryString.pick('http://example.com/?foo=1&bar=2', ['bar'])
-	//=> 'http://example.com/?bar=2'
-	```
+queryString.pick('http://example.com/?foo=1&bar=2', ['bar'])
+//=> 'http://example.com/?bar=2'
+```
 */
-export function pick(url: string, keysToPick: string[]): string;
+export function pick(url: string, keysToPick: readonly string[]): string;
 
 /**
-	Parses a query string and excludes the passed keysToExclude
+Parses a query string and excludes the passed keysToExclude
 
-	@example
-	```
-	import queryString = require('query-string');
+@example
+```
+import queryString = require('query-string');
 
-	queryString.exclude('http://example.com/?foo=1&bar=2', ['bar'])
-	//=> 'http://example.com/?foo=1'
-	```
+queryString.exclude('http://example.com/?foo=1&bar=2', ['bar'])
+//=> 'http://example.com/?foo=1'
+```
 */
-export function exclude(url: string, keysToExclude: string[]): string;
-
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -281,6 +281,32 @@ function parse(input, options) {
...
+function pick(input, keysToPick) {
+	const {url, query} = parseUrl(input);
+	const search = extract(input);
+	const getParameterOrder = key => search.indexOf(`${key}=`) === 0 ? search.indexOf(`${key}=`) : search.indexOf(`&${key}=`) === -1 ? -1 : search.indexOf(`&${key}=`);
+	const sortedKeys = Object.keys(query).sort((keyA, keyB) => getParameterOrder(keyA) - getParameterOrder(keyB));
+	const filteredParameters = sortedKeys.filter(key => keysToPick.includes(key)).map(key => `${key}=${query[key]}`);
+	const queryString = search && filteredParameters.length > 0 ? `?${filteredParameters.join('&')}` : '';
+	const hash = getHash(input);
+	return `${url}${queryString}${hash}`;
+}
```
sindresorhus, line 293: I think the logic here could be simplified by making use of the existing methods in this package.

BenoitSerrano: I agree, but couldn't find out how. Any idea for the one I could use? Maybe using an intermediate function that assembles `url` `queryString` and `hash` and that I would use in both `stringifyUrl` and `pick`?
The thing is, I can't find another function that'd preserve the parameters order :/

sindresorhus: Yes, try that. It would also nice to DRY up the repeated code in `pick` and `exclude`.

readme.md
```diff
@@ -384,6 +384,32 @@ Type: `object`
 
 Query items to add to the URL.
 
+### .pick(input, keysToPick)
+
+Returns a string with the `input` property without the keys you *did not enter* in the `keysToPick` property
+
```
sindresorhus, line 390: index.d.ts should be in sync with the readme docs-wise. That means using the same text as much as possible.

index.d.ts
```diff
@@ -327,6 +327,32 @@ Note: This behaviour can be changed with the `skipNull` option.
 */
 export function extract(url: string): string;
 
+/**
+Parses a query string and only picks the passed keysToPicks.
```
sindresorhus, line 331: I think this could be better written. And we don't need to mention parsing the query string.

sindresorhus:
Bump

sindresorhus:
Closing for lack of response.

PR #265, opened 2020-06-09
Fix sorting existing query params in URL when `sort` option is false

### LINKED ISSUE #264: Existing query params in URL get sorted even when sort is disabled for stringifyUrl
Is the following expectation correct ?

```js
test('stringify URL without sorting params', t => {
	t.deepEqual(queryString.stringifyUrl({
        url: 'https://foo.bar?C=3&A=1', query: {D: 4, B: 2}}, {sort: false}), 
             'https://foo.bar?C=3&A=1&D=4&B=2');
});
```

Currently the above test will fail because the query parameters that are already present in the URL get sorted, giving

```
'https://foo.bar?A=1&C=3&D=4&B=2'
```

dpatwardhan:
Fixes #264 .

Bug is fixed by passing the `options` object received in the `stringifyUrl` function to the internal use of `parse` function. Don't know if this could have other repercussions, but all tests passed that currently pass on master.

### DIFF
--- a/index.js
+++ b/index.js
@@ -361,7 +361,7 @@ exports.stringifyUrl = (input, options) => {
 
 	const url = removeHash(input.url).split('?')[0] || '';
 	const queryFromUrl = exports.extract(input.url);
-	const parsedQueryFromUrl = exports.parse(queryFromUrl, { sort: false });
+	const parsedQueryFromUrl = exports.parse(queryFromUrl, {sort: false});
 
 	const query = Object.assign(parsedQueryFromUrl, input.query);
 	let queryString = exports.stringify(query, options);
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -361,7 +361,7 @@ exports.stringifyUrl = (input, options) => {
 
 	const url = removeHash(input.url).split('?')[0] || '';
 	const queryFromUrl = exports.extract(input.url);
-	const parsedQueryFromUrl = exports.parse(queryFromUrl);
+	const parsedQueryFromUrl = exports.parse(queryFromUrl, options);
```
sindresorhus, line 364: I don't think `options` should be passed directly. Instead only pass the `sort` option.

PR #275, opened 2020-09-14
Add no-args case for stringify function type

mattlubner:
Hello!

I noticed that the `stringify` function's type declaration expects the first argument to always be passed. However, the actual code handles the falsy case (it returns an empty string).

I have some code which relies on that falsy case – ie. sometimes we effectively call `stringify(undefined)` – so I've updated the type declaration of `stringify` so this doesn't result in a type error.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -343,7 +343,7 @@ export interface StringifyOptions {
 Stringify an object into a query string and sort the keys.
 */
 export function stringify(
-	object: {[key: string]: any},
+	object?: {[key: string]: any},
 	options?: StringifyOptions
 ): string;
 


sindresorhus:
I don't think I want to "weaken" the types for this (and if I add it here, it has to be added to all the other methods for consistency). It was added to JS as the alternative was throwing a runtime error. But in TS, you'll get warned at compile-time. I think it makes for more readable code if you just unwrap it yourself at the callsite, instead of every API having to accept an optional type. This is the norm in other strongly-typed languages like Swift too.

If you really want to use it like this, you can just do `stringify(someOptionalObject!)`.

PR #276, opened 2020-09-25
Add support for `arrayFormat: 'bracket-separator'`

DV8FromTheWorld:
The idea behind this format is to gain the advantages of short query strings that you have with `comma` and `separator` formats while marking a field as explicitly an array via a postfixed `[]` syntax.

```js
import queryString from 'query-string'

queryString.stringify({ foo: ['one'] }, { arrayFormat: 'bracket-separator' })
// => foo[]=one

queryString.stringify({ foo: ['one', 'two', '', 'four' ] }, { arrayFormat: 'bracket-separator' })
// => foo[]=one,two,,four

queryString.stringify({ foo: ['one', 'two', '', 'four'] }, { arrayFormat: 'bracket-separator', arrayFormatSeparator: '|' })
// => foo[]=one|two||four

queryString.parse('?foo[]=one', { arrayFormat: 'bracket-separator' })
// => {foo: ['one']

queryString.parse('?foo[]=one,two,,four', { arrayFormat: 'bracket-separator' })
// => {foo: ['one', 'two', '', 'four']}

queryString.parse('?foo[]=one|two||four', { arrayFormat: 'bracket-separator', arrayFormatSeparator: '|' })
// => {foo: ['one', 'two', '', 'four']}
```

There are still are 2 question in regards to implementation:
#### What should happen for `{foo: []}`? Currently, it would drop the parameter from the stringified query which is the same as `{foo: null}`

This is undesirable personally. I feel like `{foo: []}` should produce `?foo[]` unless I specify a flag to drop it.
This would be a deviation from the `?bar&biz` that non-arrays get the treatment of when they are null (no trailing `=`), but I feel like it makes sense as we have enough information to say, explicitly, that this is an array. Whereas if it were `{foo: null}` we don't know, explicitly, that this is an array, so it would produce `?foo` which would mean that it would roundtrip correctly.
However, the only way to produce this would be to change how the system is using Array.prototype.reduce as the reducer function is not ever invoked when the array is empty.

#### What should happen for `{foo: ['one', null, 'three']}`. Currently it will drop the null and produce `foo[]=one,three`

~I haven't a strong opinion on this as it matches the functionality of `separator`. The only deviation from the `separator` functionality in regards to value dropping is that `bracket-separator` _does not_ drop empty strings.~

**[Edit]**: This has been changed to follow the `skipNull` and `skipEmptyString` options. As such, by default, `null` and `''` will not be dropped. See https://github.com/sindresorhus/query-string/pull/304 for details.
Thus, the following are true:
| Input | Options | Output |
| ------------- | ------------- | ------------ |
| `{foo: ['one', '', null, 'four']}`  | _default settings_  | `foo[]=one,,,four` |
| `{foo: ['one', '', null, 'four']}`  | `skipNull: true`  |  `foo[]=one,,four` |
| `{foo: ['one', '', null, 'four']}`  | `skipEmptyString true`  |  `foo[]=one,,four` |
| `{foo: ['one', '', null, 'four']}`  | `skipEmptyString true, skipNull: true`  |  `foo[]=one,four` |


sindresorhus:
> What should happen for {foo: []}?

`?foo[]` would indeed be the best output.

> What should happen for {foo: ['one', null, 'three']}

I think dropping the `null` is the best option.

sindresorhus:
You need to explicitly document the behavior with empty array, null in array, and empty string in array in the docs.

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+.idea/
```
sindresorhus, line 3: This should be in your own global gitignore, not here.

index.d.ts
```diff
@@ -45,6 +45,18 @@ export interface ParseOptions {
...
 
+	- `bracket-separator`: Parse arrays (that are explicitly marked with brackets) with elements separated by a custom character:
+
+		```
+		import queryString = require('query-string');
+
+		queryString.parse('foo[]=1', {arrayFormat: 'bracket-separator', arrayFormatSeparator: '|'});
+	 	//=> {foo: ['1']}
+
+		queryString.parse('foo[]=1|2|3', {arrayFormat: 'separator', arrayFormatSeparator: '|'});
```
sindresorhus, line 56: It would be more valuable to show 
```suggestion
		queryString.parse('foo[]=1', {arrayFormat: 'separator', arrayFormatSeparator: '|'});
```

So the user could also see the difference.

index.d.ts
```diff
@@ -45,6 +45,18 @@ export interface ParseOptions {
 		//=> {foo: ['1', '2', '3']}
 		```
 
+	- `bracket-separator`: Parse arrays (that are explicitly marked with brackets) with elements separated by a custom character:
+
+		```
+		import queryString = require('query-string');
+
+		queryString.parse('foo[]=1', {arrayFormat: 'bracket-separator', arrayFormatSeparator: '|'});
```
sindresorhus, line 53: Add an example for this with multiple values too.

sindresorhus:
I would also like to see more tests.

sindresorhus (changes requested):

sindresorhus:
And sorry for the slow reply. I've just been overwhelmed with notifications.

index.d.ts
DV8FromTheWorld: I don't quite follow what this suggestion is recommended. Doesn't this suggestion modify the documentation to be the same as the other `bracket-separator` example?

As a note, I have just noticed that this second example incorrectly shows `arrayFormat: 'separator'` when it should be `arrayFormat: 'bracket-separator'`

DV8FromTheWorld: I'm not quite following what this is wanting.

 I did realize I accidently left the `arrayFormat` set to `separator` here instead of `bracket-separator`, so I've updated that as needed.

DV8FromTheWorld:
I believe I've covered all that was asked. Please let me know if something else needs attention.
Thanks!

sindresorhus (changes requested):

sindresorhus:
Bump

DV8FromTheWorld:
Still have eyes on this, just haven't had a second to address. Will l fix this up soon

index.js
```diff
@@ -53,8 +53,29 @@ function encoderForArrayFormat(options) {
 					return result;
 				}
 
+				const keyValueSep = options.arrayFormat === 'bracket-separator' ?
```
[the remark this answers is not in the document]
DV8FromTheWorld: Yup. It lingers from when I was originally working on the `separator` clause. You can actually seem the code we duplicated erroneously in the `bracket-separator` clause as well.

As I have now remerged these clauses to cut down on code duplication, the above should be addressed.

DV8FromTheWorld:
Review comments have been addressed. Thank you for your patience.

sindresorhus:
Can you fix the merge conflict?

index.js
```diff
@@ -127,6 +139,28 @@ function parserForArrayFormat(options) {
 				accumulator[key] = newValue;
 			};
 
+		case 'bracket-separator':
+			return (key, value, accumulator) => {
+				const isArray = /(\[\])$/.exec(key);
```
sindresorhus, line 144: This should use `.test()`

index.js
```diff
@@ -127,6 +139,28 @@ function parserForArrayFormat(options) {
...
+				key = key.replace(/\[\]$/, '');
+
+				if (!isArray) {
+					accumulator[key] = value ? decode(value, options) : value;
+					return;
+				}
+
+				const arrValue = value === null ?
```
sindresorhus, line 152:
```suggestion
				const arrayValue = value === null ?
```

DV8FromTheWorld:
Merge conflicts resolved.
Documentation updated to reflect new `skipNull` and `skipEmptyString` changes.
PR suggestions included.

PR #282, opened 2020-10-03
Add `.pick()` and `.exclude()`

Richienb:
Fixes: #259


index.d.ts
```diff
@@ -404,3 +404,40 @@ export function stringifyUrl(
 	object: UrlObject,
 	options?: StringifyOptions
 ): string;
+
+/**
+Filter query parameters from a url.
+
+@param url The url containing the query parameters to filter.
+@param includeParams The names of the query parameters to retain. All other query parameters will be removed from the url.
```
sindresorhus, line 412:
```suggestion
@param keysToKeep The names of the query parameters to retain. All other query parameters will be removed from the URL.
```

?

I think that name might be clearer. The description needs to be updated for that too.

index.d.ts
```diff
@@ -404,3 +404,40 @@ export function stringifyUrl(
...
+queryString.filterElements('https://foo.bar?foo=1&bar=2#hello', ['foo']);
+//=> 'https://foo.bar?foo=1#hello'
+
+queryString.filterElements('https://foo.bar?foo=1&bar=2#hello', (name, value) => value === 2, {parseNumbers: true});
+//=> 'https://foo.bar?bar=2#hello'
+```
+*/
+export function filterElements(
+	url: string,
+	includeParams: ReadonlyArray<string>,
```
sindresorhus, line 426:
```suggestion
	includeParams: readonly string[],
```

sindresorhus:
What's the reasoning for going with a single method instead of pick/exclude? Just curious. And why `.filterElements()` over `.filter()`?

readme.md
```diff
@@ -415,6 +415,43 @@ Type: `object`
...
+
+#### includeParams
+
+Type: `string[]`
+
+The names of the query parameters to retain. All other query parameters will be removed from the url.
+
+#### filter
+
+Type: `(name, value) => boolean`
```
sindresorhus, line 445: I would use the name `key` over `name`. Applies elsewhere too.

sindresorhus (changes requested):

Richienb:
> What's the reasoning for going with a single method instead of pick/exclude? Just curious. And why .filterElements() over .filter()?

I took inspiration for naming from https://github.com/sindresorhus/query-string/issues/259#issuecomment-629611229

index.d.ts
```diff
@@ -404,3 +404,41 @@ export function stringifyUrl(
...
 	options?: StringifyOptions
 ): string;
+
+/**
+Filter query parameters from a URL.
+
+@param url - The URL containing the query parameters to filter.
+@param keysToKeep - The names of the query parameters to keep. All other query parameters will be removed from the URL.
+@param filter - A filter predicate that will be provided the name of each query parameter and its value. The `parseNumbers` and `parseBooleans` options also affect `value`.
+@returns The URL with the specified query parameters removed.
```
sindresorhus, line 414: This is the inverse of what's happening.

sindresorhus:
> I took inspiration for naming from #259 (comment)

That was a response to other naming suggestions. I think just `.filter()` is clear enough in this case.

index.d.ts
```diff
@@ -404,3 +404,41 @@ export function stringifyUrl(
 	object: UrlObject,
 	options?: StringifyOptions
 ): string;
+
+/**
+Filter query parameters from a URL.
```
sindresorhus, line 409: This description about **removing** query parameters, while the below arguments is about **keeping**.

index.d.ts
```diff
@@ -404,3 +404,41 @@ export function stringifyUrl(
...
+
+@param url - The URL containing the query parameters to filter.
+@param keysToKeep - The names of the query parameters to keep. All other query parameters will be removed from the URL.
+@param filter - A filter predicate that will be provided the name of each query parameter and its value. The `parseNumbers` and `parseBooleans` options also affect `value`.
+@returns The URL with the specified query parameters removed.
+
+@example
+```
+queryString.filterElements('https://foo.bar?foo=1&bar=2#hello', ['foo']);
```
sindresorhus, line 418: I think the array version should support exclusions too by doing `['foo', '!bar']`

sindresorhus:
> For a better implementation, this could preserve the order of url queries, cause input and output are both strings, which has order. - https://github.com/sindresorhus/query-string/issues/259#issuecomment-632185453

Can you add a test for that?

index.d.ts
Richienb: By *filter*, I mean more like *filter out* except that wording might implicitly mean *take out* in this context. What do you think the description should be?

index.d.ts
Richienb: We should pass that to https://github.com/sindresorhus/map-obj

Richienb:
> Can you add a test for that?

If you mean the order of query parameters, that is already tested.

index.d.ts
Richienb: If exclusions and inclusions are used in the same array, it presents a different problem. What would `['foo', '!bar']` do vs `['foo']`? Both would produce the same object.

Richienb: Perhaps we could instead add `.filter.exclude()` which does the same as `.filter()` except the other way round.

sindresorhus: Which is why the original suggestion was to add `.pick()` and `.exclude()` methods, which I still prefer.

Richienb: https://github.com/sindresorhus/filter-obj/issues/9

sindresorhus:
> Which is why the original suggestion was to add .pick() and .exclude() methods, which I still prefer.

@Richienb Any thoughts on this? I think that might be a better API for this.

Richienb:
@sindresorhus I am waiting on https://github.com/sindresorhus/filter-obj/issues/9 to implement this.

sindresorhus:
> I am waiting on sindresorhus/filter-obj#9 to implement this.

I don't have time to do https://github.com/sindresorhus/filter-obj/issues/9 and doesn't look like anyone else has either, so should we close this PR for now? It's stale and doesn't seem like something will happen here anytime soon.

Richienb:
@sindresorhus I guess I'll tackle the issue myself.

Since specifying keys to be included and excluded is mutually exclusive, should I add it as a separate option in `filter-obj` instead of exclamation marks?

sindresorhus:
Sure

index.d.ts
```diff
@@ -404,3 +404,79 @@ export function stringifyUrl(
...
 	options?: StringifyOptions
 ): string;
+
+/**
+Pick query parameters from a URL.
+
+@param url - The URL containing the query parameters to pick.
+@param keysToKeep - The names of the query parameters to keep. All other query parameters will be removed from the URL.
+@param filter - A filter predicate that will be provided the name of each query parameter and its value. The `parseNumbers` and `parseBooleans` options also affect `value`.
+@returns The URL with the query parameters picked.
```
sindresorhus, line 414:
```suggestion
@returns The URL with the picked query parameters.
```

readme.md
```diff
@@ -415,6 +415,53 @@ Type: `object`
...
+//=> 'https://foo.bar?bar=2#hello'
+```
+
+#### url
+
+Type: `string`
+
+The URL containing the query parameters to filter.
+
+#### keysToKeep | keysToRemove
```
sindresorhus, line 447: These should be separate entries.

sindresorhus (changes requested):

readme.md
```diff
@@ -415,6 +415,52 @@ Type: `object`
...
+
+### .exclude(url, keys, options?)
+### .exclude(url, filter, options?)
+
+Exclude query parameters from a URL.
+
+Returns a string with the new URL.
+
+```js
+queryString.filter('https://foo.bar?foo=1&bar=2#hello', ['foo']);
```
sindresorhus, line 433: Outdated example.

readme.md
```diff
@@ -415,6 +415,52 @@ Type: `object`
 
 Query items to add to the URL.
 
+### .pick(url, keys, options?)
+### .pick(url, filter, options?)
+
+Pick query parameters from a URL.
+
+Returns a string with the new URL.
+
```
sindresorhus, line 424: Add code block like in index.d.ts

readme.md
```diff
@@ -415,6 +415,52 @@ Type: `object`
...
+Type: `string`
+
+The URL containing the query parameters to filter.
+
+#### keys
+
+Type: `string[]`
+
+The names of the query parameters to keep or remove depending on the function used. All other query parameters will be removed from the URL.
```
sindresorhus, line 450: > All other query parameters will be removed from the URL.

This is confusing when it describes `exclude`. Because it makes it sound like there are no query parameters left.

index.js
```diff
@@ -376,3 +377,18 @@ exports.stringifyUrl = (input, options) => {
...
+
+	const {url, query, fragmentIdentifier} = exports.parseUrl(input, options);
+	return exports.stringifyUrl({
+		url,
+		query: filterObject(query, filter),
+		fragmentIdentifier
+	}, options);
+};
+
+exports.exclude = (input, filter, options) => exports.pick(input, Array.isArray(filter) ? key => !filter.includes(key) : (key, value) => !filter(key, value), options);
```
sindresorhus, line 394: This is pretty hard to read. Use some more lines and make it more readable.

sindresorhus (changes requested):

PR #284, opened 2020-10-03
Add tests to confirm existing number parsing behavior

### LINKED ISSUE #263: Improve number parsing
https://github.com/sindresorhus/query-string/pull/241

RocktimSaikia:
Added some extra tests to confirm the existing behavior of number parsing of `query-string`.
fixes #263 
// @sindresorhus

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -75,7 +75,7 @@ test('handle `+` correctly', t => {
 	t.deepEqual(queryString.parse('foo+faz=bar+baz++'), {'foo faz': 'bar baz  '});
 });
 
-test('handle number pasring correctly with math operators', t => {
+test('additional tests to clarify exisiting number parsing behavior', t => {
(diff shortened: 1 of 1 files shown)


test/parse.js
```diff
@@ -75,6 +75,11 @@ test('handle `+` correctly', t => {
 	t.deepEqual(queryString.parse('foo+faz=bar+baz++'), {'foo faz': 'bar baz  '});
 });
 
+test('additional tests to clarify exisiting number parsing behavior', t => {
```
sindresorhus, line 78: The test should have a description reflecting what it's testing, which is "parses numbers with exponential notation as string"

RocktimSaikia: @sindresorhus Ah right. Thanks for the suggestion. I always have a hard time naming variables and test descriptions. May my I think about it too much. Anyway, I updated with the description :+1:

sindresorhus:
Thanks :)

PR #287, opened 2020-10-15
When only receiving values with encoded array value, decode values

mishugana:
Fixes https://github.com/sindresorhus/query-string/issues/242
Now all 99 tests pass (instead of one failing on purpose)
when pasting my url that used a custom array separator into slack, it was uri encoded, which broke things.
now in situations like that, the simple case will work, but allows more complicated cases to also still work.

### DIFF
--- a/index.js
+++ b/index.js
@@ -123,7 +123,9 @@ function parserForArrayFormat(options) {
 		case 'separator':
 			return (key, value, accumulator) => {
 				const isArray = typeof value === 'string' && value.split('').indexOf(options.arrayFormatSeparator) > -1;
-				const newValue = isArray ? value.split(options.arrayFormatSeparator).map(item => decode(item, options)) : value === null ? value : decode(value, options);
+				const isEncodedArray = (typeof value === 'string' && !isArray && decode(value, options).split('').indexOf(options.arrayFormatSeparator) > -1);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -123,7 +123,9 @@ function parserForArrayFormat(options) {
 		case 'separator':
 			return (key, value, accumulator) => {
 				const isArray = typeof value === 'string' && value.split('').indexOf(options.arrayFormatSeparator) > -1;
-				const newValue = isArray ? value.split(options.arrayFormatSeparator).map(item => decode(item, options)) : value === null ? value : decode(value, options);
+				const isEncodedArray = (typeof value === 'string' && !isArray && decode(value, options).split('').indexOf(options.arrayFormatSeparator) > -1);
```
sindresorhus, line 126: I don't see the point in splitting the string into an array. You can just look for the character in the string. Also use `Array#includes` instead.

mishugana:
I changed over the code to use the includes method. All tests still pass. If you think this is an ok pull request, and you dont mind, would you be able to tag this pull request with hacktoberfest-accepted (if it does get accepted) since I am trying to get one of those free t-shirts.

sindresorhus:
I assume merging will do the same.

PR #290, opened 2020-11-04
Added Bundling For Browsers And Other Module Types

muratgozel:
Hi,

The library wasn't compatible to be included via HTML script tags. I don't really know what is the maintainers opinion about this but I've implemented a code bundling solution and started using it in my project.

- Bundling done by rollup.js and babel.
- The maintainer needs to run `npm run build` on each new release.
- Code bundles are placed in `dist` folder. There are different types of bundles but the one `dist/query-string.iife.js` can be used within the script tag.
- Browser bundle exports a global variable called `QueryString` so the library can be accessed with `window.QueryString` and continues to work as usual.


sindresorhus:
Sorry, I'm not interested in bundling. I would recommend to publish a fork.

PR #291, opened 2020-11-04
Allow readonly `Stringifiable[]` in `StringifiableRecord`

tpict:
Since the first argument to `stringify` isn't mutated, readonly arrays should be allowed as record values. This is a backwards-compatible change, you'll still be able to pass a mutable array if you like.


tpict:
thank you!

PR #306, opened 2021-02-03
Ignore ending ampersand when parsing

pytnik23:
Hi there!

Sometimes query string has an extra ampersand in the end. I think it shouldn't be in the result. What do you think about it?

**String to parse**
`foo=bar&`

**Actual result**
`{ '': null, foo: 'bar' }`

**Expected result**
`{ foo: 'bar' }`

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -15,6 +15,7 @@ test('query strings starting with a `&`', t => {
 
 test('query strings ending with a `&`', t => {
 	t.deepEqual(queryString.parse('foo=bar&'), {foo: 'bar'});
+	t.deepEqual(queryString.parse('foo=bar&&&'), {foo: 'bar'});
 });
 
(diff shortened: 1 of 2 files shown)


test/parse.js
```diff
@@ -13,6 +13,10 @@ test('query strings starting with a `&`', t => {
 	t.deepEqual(queryString.parse('&foo=bar&foo=baz'), {foo: ['bar', 'baz']});
 });
 
+test('query strings ending with a `&`', t => {
+	t.deepEqual(queryString.parse('foo=bar&'), {foo: 'bar'});
```
sindresorhus, line 17: Can you also do a test for `foo=bar&&&`?

pytnik23: Sure, good idea. It's added

PR #316, opened 2021-05-28
Stick to ES5

tjenkinson:
We'd like to use this lib in an old environment which is only guaranteed to have ES5 support.

Things like rest arguments are fine because we can transpile those, but we can't transpile the changes in here.

WDYT?

Would be great if we don't need to maintain a fork for our use case.

Thanks


sindresorhus:
Sorry, that's not going to happen. You can either transpire with Babel or use version 5.

Issue #296, opened 2020-11-17
Parsing of `arrayFormat: 'comma'` not working for queries of single value with comma


marxlow:
Say we have an Object we are stringifying to become a query param of the following shape (notice the comma in the value)

1. Create an object with its value as a single string / array of one string.

```
const singleQueryObject = { not_important: [ "I, am, one, single, value"]};

// OR (the value can also just be a plain string, not an array like the above)

const singleQueryObject = { not_important: "I, am, one, single, value" };
```

2. We stringify it:

```
const stringifyResult = queryString.stringify(singleQueryObject, { arrayFormat: 'comma' });
//=>  'not_important=I%2C%20am%2C%20one%2C%20single%2C%20value'
```

3. We parse it (Actual)

```
const parsedResult = queryString.parse(stringifyResult, { arrayFormat: 'comma' });
//=> { not_important: [ 'I', ' am', ' one', ' single', ' value' ] }
```

(Expected)
```
{ not_important: [ 'I, am, one, single, value' ] };
```

## Work around (?)
This bug only occurs when there is a **single value**. So say if we run the same step as above with this object instead:

```
const singleQueryObject = { not_important: [ 'I, am, one, single, value', 'It works!'] };
const stringifyResult = queryString.stringify(singleQueryObject, { arrayFormat: 'comma' });
const parsedResult = queryString.parse(stringifyResult, { arrayFormat: 'comma' });

//=> { not_important: [ 'I, am, one, single, value', 'It works!' ] }
```

Yashkochar20:
@sindresorhus @Richienb This is also causing issues on our side ? Any timeline for this fix ?

sindresorhus:
// @mishugana

Yashkochar20:
@sindresorhus @mishugana any update on this ?

decimoseptimo:
```js
  console.log(
    queryString.stringify(
      { a: ["asd", null, "123", ""] },
      {
        arrayFormat: "index",
        sort: false,
      }
    )
  )
//a[0]=asd&a[1]&a[2]=123&a[3]=

 console.log(
      queryString.stringify(
        { a: ["asd", null, "123", ""] },
        {
          arrayFormat: "comma",
          sort: false,
        }
      )
  )
//a=asd,123
```
`arrayFormat: "comma"`  isn't working as we're expecting `a=asd,,123,`

cainlevy:
The current `parse` API doesn't appear to have enough information to solve this problem.

[...]

PR #322, opened 2021-06-24
Update index.js

Inderpreetsingh1064:
Fixed the issue #296 , by modifying and adding to the code to deal with one particular case causing problem.


sindresorhus:
This is not something that can be merged. If you have a proposal on how to resolve #296, I suggest you describe it in words in the issue instead.

PR #326, opened 2021-08-19
breaking: publish esm (fixes #352)

aladdin-add:
the main changes are:
* drop node < 12
* convert index.js => esm
* pkg.type => "module"
* upgrade to ava v3
* update examples in docs

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
 	"engines": {
 		"node": ">=12"
 	},
+	"type": "module",
 	"scripts": {
 		"benchmark": "node benchmark.js",
 		"test": "xo && ava && tsd"
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
 		"node": ">=12"
 	},
 	"type": "module",
+	"exports": "./index.js",
 	"scripts": {
 		"benchmark": "node benchmark.js",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -50,5 +51,11 @@
 		"fast-check": "^1.5.0",
 		"tsd": "^0.7.3",
 		"xo": "^0.24.0"
+	},
+	"xo": {
+		"rules": {
+			"unicorn/import-index": 0,
+			"import/extensions": 0
+		}
```
aladdin-add, line 59: this is required as ava does not fully support esm.

```bash
"Error [ERR_UNSUPPORTED_DIR_IMPORT]"
```

you can remove it after ava esm supports.

sindresorhus:
I appreciate the PR, but I don't intend to convert this package u til later this year (it will cause too much support). I also prefer to do it myself as it will take much more work to comment on what needs changing than just doing the conversion myself.

aladdin-add:
do you mind add `"type": "commonjs"` or `"main": "index.js"`, to help tools better understanding it's a cjs module?

sindresorhus:
No, tools are broken if they don't already assume it's CJS when there's to "type:module".

aladdin-add:
well, it also makes sense. 👍

PR #352, opened 2022-12-01
Allow multi-charactor separators

kstenerson-ab:
There is nothing in the implementation that breaks with multi-character separators,
so the single character limitation is a bit arbitrary.

A system I work on uses `;;` as the separator in its HTTP API.


sindresorhus:
You need to update the docs (readme and TS doc comments) to mention that multiple characters are allowed.

PR #361, opened 2022-12-13
Update index.js

gangsthub:
This is what is specified [in the .d.ts file](https://github.com/sindresorhus/query-string/blob/main/index.d.ts#L4-L6), and apart from that, it's breaking the import if you just `export * as default`. (Or I couldn't make it work!)


sindresorhus:
The existing syntax should be valid, but I'll merge as I don't have time to deal with support.

PR #365, opened 2022-12-19
Upgrade dependencies

### LINKED ISSUE #359: Decode-uri-component requirement forces CommonJS
Decode-uri-component is CommonJS, which means I can't use it (webpack converts it, but web-dev-server which we use doesn't support it out of the box, and the rollup plugin which would add conversion is horribly broken in it).
Can decode-uri-component be made inline or such?
Or can we have a -lite version which only does stringify (I only need that), and which doesn't require Decode-uri-component ? :)

SamVerschueren:
fixes #359 

This PR upgrades the dependencies of `decode-uri-component` and `split-on-first`. They both switched to ESM. I also included type definitions in the latest version of `decode-uri-component`.

### DIFF
--- a/base.js
+++ b/base.js
@@ -353,7 +353,13 @@ export function parse(query, options) {
 			continue;
 		}
 
-		let [key, value] = splitOnFirst(options.decode ? parameter.replace(/\+/g, ' ') : parameter, '=');
+		const param = options.decode ? parameter.replace(/\+/g, ' ') : parameter;
+
+		let [key, value] = splitOnFirst(param, '=');
+
+		if (key === undefined) {
+			key = param;
+		}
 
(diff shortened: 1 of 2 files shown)


base.js
```diff
@@ -353,7 +353,13 @@ export function parse(query, options) {
 			continue;
 		}
 
-		let [key, value] = splitOnFirst(options.decode ? parameter.replace(/\+/g, ' ') : parameter, '=');
+		const param = options.decode ? parameter.replace(/\+/g, ' ') : parameter;
+
+		let [key, value] = splitOnFirst(param, '=');
```
SamVerschueren, line 358: `splitOnFirst()` now returns an empty array if it couldn't find the separator.

PR #249, opened 2020-03-02
fix issue 210: custom values formatting schema

goodwin64:
Resolves https://github.com/sindresorhus/query-string/issues/210

PR #369, opened 2023-08-05
feat: parseNumbers support includes & excludes options

daiwa233:
1. Added logic to return a string when the value exceeds the maximum safe JavaScript integer
2. Gives the user more flexibility to customize what to parse or not to parse when setting up parseNumbers.

### DIFF
--- a/base.d.ts
+++ b/base.d.ts
@@ -127,6 +127,7 @@ export type ParseOptions = {
 
 	/**
 	Parse the value as a number type instead of string type if it's a number.
+    Note: will always return a string if it exceeds the maximum safe integer in JavaScript
 
 	@default false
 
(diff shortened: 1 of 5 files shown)


daiwa233:
1. Added logic to return a string when the value exceeds the maximum safe JavaScript integer
2. Gives the user more flexibility to customize what to parse or not to parse when setting up parseNumbers.

base.d.ts
```diff
@@ -136,9 +137,15 @@ export type ParseOptions = {
...
+
+	queryString.parse('foo=1&bar=2', {parseNumbers: {includes: ['foo']}});
+	//=> {foo: 1, bar: '2'}
+
+	queryString.parse('foo=1&bar=2', {parseNumbers: {excludes: ['foo']}});
+	//=> {foo: '1', bar: 2}
 	```
 	*/
-	readonly parseNumbers?: boolean;
+	readonly parseNumbers?: boolean | {includes?: string[]; excludes?: string[]};
```
sindresorhus, line 148: This feels too limiting. I would prefer a filter method that receives the key and value and returns true whether it should be parsed as a number. But the ideal solution would be something like https://github.com/sindresorhus/query-string/issues/210

daiwa233: Got it. I'll look at #249 remaining unfinished parts

sindresorhus: If you decide to work no #249, I would recommend posting a proposal on what to do in #249 before writing any code to make sure it's the most optimal solution.

PR #377, opened 2024-01-30
recognize bigint as stringifiable

bug-brain:
This would be a nice feature to prevent some explicit `.toString()´ calls before passing an object to `stringify`.

### DIFF
--- a/base.d.ts
+++ b/base.d.ts
@@ -412,7 +412,7 @@ export type StringifyOptions = {
 	readonly skipEmptyString?: boolean;
 };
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
I'm willing to merge this, but you need to fix the lint issues.

bug-brain:
Going for xo@0.56.0 would require using `String#replaceAll()` instead of `String#replace()`. For the two usages this would not be breaking but I prefer changing less for this feature because it is technically only a typing issue.

PR #392, opened 2024-09-19
FIX: Decode back-separator value before splitting it with separator

### LINKED ISSUE #388: Bug: comma separated arrays not parsed correctly from encoded url
Test case:

```ts
it('should parse arrays correctly', () => {
    const search =
      '?testA%5B%5D=1&testB%5B%5D=a%2Cb%2Cc%2Cd%2Ce%2Cf&testC=true';

    const result = queryString.parse(search, {
      parseBooleans: true,
      parseNumbers: true,
      arrayFormat: 'bracket-separator',
    });

    expect(result).toEqual({
      testA: [1],
      testB: ['a', 'b', 'c', 'd', 'e', 'f'],
      testC: true,
    });
  });
```

results in:

image

when it's unencoded it works fine:

```ts
it('should parse arrays correctly', () => {
    const search =
      '?testA%5B%5D=1&testB%5B%5D=a,b,c,d,e,f&testC=true';

    const result = queryString.parse(search, {
      parseBooleans: true,
      parseNumbers: true,
      arrayFormat: 'bracket-separator',
    });

    expect(result).toEqual({
      testA: [1],
      testB: ['a', 'b', 'c', 'd', 'e', 'f'],
      testC: true,
    });
  });
```

Library version: 9.1.0

scottenock:
The original implementation was splitting the received value before it then decoded the value. Because the value was URL encoded, the split function could not find the separator to successfully split. Adjusted the order of operations which fixes #388

- Adjusted order of operations, ensuring we decode before splitting the value
- Implemented test case to guard against this regression.

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -220,8 +220,8 @@ test('query strings having a brackets+separator array and format option as `brac
 		arrayFormat: 'bracket-separator',
 	}), {
 		testA: ['1'],
-      	testB: ['a', 'b', 'c', 'd', 'e', 'f'],
-      	testC: 'true',
+		testB: ['a', 'b', 'c', 'd', 'e', 'f'],
+		testC: 'true',
 	});
 });
 
--- a/test/parse.js
+++ b/test/parse.js
@@ -216,12 +216,12 @@ test('query strings having a brackets+separator array and format option as `brac
 });
 
 test('query strings having a brackets+separator array and format option as `bracket-separator` with a URL encoded value', t => {
-	t.deepEqual(queryString.parse('?testA%5B%5D=1&testB%5B%5D=a%2Cb%2Cc%2Cd%2Ce%2Cf&testC=true', {
+	const key = 'foo[]';
+	const value = 'a,b,c,d,e,f';
+	t.deepEqual(queryString.parse(`?${encodeURIComponent(key)}=${encodeURIComponent(value)}`, {
 		arrayFormat: 'bracket-separator',
 	}), {
-		testA: ['1'],
-		testB: ['a', 'b', 'c', 'd', 'e', 'f'],
-		testC: 'true',
+		foo: ['a', 'b', 'c', 'd', 'e', 'f'],
 	});
 });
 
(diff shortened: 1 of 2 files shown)


test/parse.js
```diff
@@ -215,6 +215,16 @@ test('query strings having a brackets+separator array and format option as `brac
 	}), {foo: ['']});
 });
 
+test('query strings having a brackets+separator array and format option as `bracket-separator` with a URL encoded value', t => {
+	t.deepEqual(queryString.parse('?testA%5B%5D=1&testB%5B%5D=a%2Cb%2Cc%2Cd%2Ce%2Cf&testC=true', {
```
sindresorhus, line 219: I would prefer the string to be unencoded, and encode it before passing it to `parse`. This makes it possible to actually see what it contains when reading the code.

test/parse.js
```diff
@@ -215,6 +215,16 @@ test('query strings having a brackets+separator array and format option as `brac
 	}), {foo: ['']});
 });
 
+test('query strings having a brackets+separator array and format option as `bracket-separator` with a URL encoded value', t => {
+	const key = 'foo[]';
+	const value = 'a,b,c,d,e,f'
+	t.deepEqual(queryString.parse(`?${encodeURIComponent(key)}=${encodeURIComponent(value)}`, {
```
scottenock, line 221: NOTE: I have had to encode the key and value separately as the `encodeURIComponent` function will encode the `?` and `=` characters which breaks the test. 

I attempted `encodeURI` function but this doesn't encode the `,` character which we need encoded for this test.

scottenock:
Thanks for reviewing @sindresorhus 🙌 

I've adjusted the test to be more readable with regards to the encoded strings. There's some existing tests that are using encoded values without showing the un-encoded version. Do you want me to adjust these to conform to your comment?

sindresorhus:
> Do you want me to adjust these to conform to your comment?

Yeah, that would be great.

scottenock:
Sweet @sindresorhus, I've updated the tests and they should be more readable. Let me know if there's any issues.

sindresorhus:
Thanks :)

PR #395, opened 2024-10-24
Fix bug

OmarMohamedoi:
In this fix the solution handles null and empty strings properly in the stringify function when using arrayFormat:'comma'. The fix ensures that null values and empty strings are not skipped when the skipNull and skipEmptyString options are set to false. it addresses how the comma array format should properly place commas for those null and empty string values in the output.

### DIFF
--- a/.js
+++ b/.js
@@ -1,0 +1,19 @@
+const queryString = require('./index');
+
+// Example data that includes null and empty strings
+const params = {
+    list: ['item', '', null, 'last']
+};
+
+// Options to reproduce the bug
+const options = {
+    arrayFormat: 'comma',
+    skipNull: false,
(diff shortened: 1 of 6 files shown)


.js
```diff
@@ -0,0 +1,19 @@
...
+    arrayFormat: 'comma',
+    skipNull: false,
+    skipEmptyString: false
+};
+
+// Stringify the parameters with the options
+const result = queryString.stringify(params, options);
+
+// Log the result to console
+console.log(result); // Expected to incorrectly skip null and empty strings based on the bug
```
scottenock, line 19: This example should be in the comment of the PR or expressed via a passing unit test. This file can't be merged into main

PR #400, opened 2024-12-05
Add export * from 'base.js' for better modularity

LeeSamFong:
I hope to add this codes to support friendlier tree-shaking

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,6 +3,7 @@
 // Workaround for TS missing feature.
 import * as queryString from './base.js';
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Tree-shaking should work fine even without this. If not, the problem is the bundler you use.

LeeSamFong:
Thank you, sindresorhus, for your reply.

LeeSamFong:
If I have any misunderstandings, can sindresorhus point them out for me?

PR #407, opened 2025-05-08
FEAT: implement support for boolean type

scottenock:
Hey @sindresorhus , when implementing the parsing fix for PR https://github.com/sindresorhus/query-string/pull/406 I noticed it would be pretty quick to add support for a `boolean` type. This allows the user to override the behaviour of `parseBoolean`. 

I've updated the `README` and type definitions to include the new `boolean` type.

I've also updated the description of the types feature to be clearer. Let me know your thoughts!

### DIFF
--- a/base.d.ts
+++ b/base.d.ts
@@ -276,15 +276,17 @@ export type ParseOptions = {
 
 	@example
 	Force `flagged` to be parsed as a boolean even when `parseBooleans` is false:
-	````
-	queryString.parse('?isAdmin=true&flagged=true', {
+	```
+	queryString.parse('?isAdmin=true&flagged=true&isOkay=0', {
 			parseBooleans: false,
 			types: {
 					flagged: 'boolean',
+					isOkay: 'boolean',
 			},
 	});
-	//=> { isAdmin: 'true', flagged: true }
-	````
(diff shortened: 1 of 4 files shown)


sindresorhus:
Thanks for implementing this. I think it should also accept `0` and `1` as those are commonly used as booleans.

scottenock:
Good shout! I've added support for `0` and `1` when the boolean type is used.

I haven’t extended this to the parseBoolean option yet—wanted to check with you first @sindresorhus . If you think it makes sense, I’ll get it added. Otherwise, I believe the PR is good to go 💪

sindresorhus:
> I haven’t extended this to the parseBoolean option yet—wanted to check with you first @sindresorhus . If you think it makes sense, I’ll get it added. Otherwise, I believe the PR is good to go 💪

No, it should not. `parseBooleans` means that it would parse numbers as booleans, even if it's meant to be a number. For example, a query parameter that could be 0, 1, 2, 3, etc.

sindresorhus:
Yeah, that makes sense to me.

scottenock:
@sindresorhus I'll get this implemented :muscle:

